PR #28, opened 2023-09-28
EW-502 provided a get organization by ID API

Fshmit:
# Description
A get organization by ID API was implemented.

https://ticketsystem.dbildungscloud.de/browse/EW-502

## Approval for review


src/modules/organisation/domain/organisation.service.ts
```diff
@@ -17,4 +17,12 @@ export class OrganisationService {
         }
         return { ok: false, error: new EntityCouldNotBeCreated(`Organization could not be created`) };
     }
+
+    public async findOrganisationById(id: string): Promise<Result<OrganisationDo<true>, DomainError>> {
+        const organisation: Option<OrganisationDo<true>> = await this.organisationRepo.findById(id);
+        if (organisation) {
+            return { ok: true, value: organisation };
+        }
+        return { ok: false, error: new EntityNotFoundError(`Organization with the following ID ${id} does not exist`) };
```
psachmann, line 26: Maybe we could the message for the entity not found error more generic? So that not every caller has a different message.

src/modules/organisation/domain/organisation.service.spec.ts
```diff
@@ -69,4 +70,34 @@ describe('OrganisationService', () => {
...
+                organisationDo.id,
+            );
+            expect(result).toEqual<Result<OrganisationDo<true>>>({
+                ok: true,
+                value: organisationDo as unknown as OrganisationDo<true>,
+            });
+        });
+
+        it('should return a domain error', async () => {
+            const organisationDo: OrganisationDo<false> = DoFactory.createOrganisation(false);
```
psachmann, line 89: As previous mentioned, we can generate an id over the `DoFactory`.

src/modules/organisation/domain/organisation.service.spec.ts
```diff
@@ -69,4 +70,34 @@ describe('OrganisationService', () => {
...
+            organisationDo.id = faker.string.uuid();
```
psachmann, line 90: Do we need a do? Can't we just create an id with faker?

psachmann (changes requested):

src/modules/organisation/api/organisation.controller.ts
```diff
@@ -38,4 +40,22 @@ export class OrganisationController {
         const createdOrganisation: CreatedOrganisationDto = await this.uc.createOrganisation(organisationDto);
         return this.mapper.map(createdOrganisation, CreatedOrganisationDto, OrganisationResponse);
     }
+
+    @Get(':organisationId')
+    @ApiCreatedResponse({ description: 'The organization was successfully pulled.' })
```
psachmann, line 45: This should be an ApiOkResponse and the response type is not shown on the OpenAPI documentation.

psachmann (commented):
- the property value typ in OrganisationResponse is not written in all upper case in the json response
- the ticket does not specify a namensergaenzung filed in the response
- the result hadder if the organisation is not found is 200 and not 404
- the result body for a 404 request does not match the body specified in the ticket

psachmann (approved):

PR #50, opened 2023-10-20
feature/spsh-130_kc_health_probe

kristoff-kiefer:
# Description

Introduced a health check to make sure the keycloak is up and the backend correctly configured to get a client (Authentication)

## Changes

Adds a new health probe so that we can be sure that a Kubernetes deployment actually works.

## Approval for review
- [x] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.


alweber-cap (changes requested):
Functionally good, although a few architectural comments:
- KeycloakHealthIndicator returns directly a Controller Result. Even though it comes from a nest addon, i feel that we are ignoring Adapter/Domain borders here
- LoginService is currently in ui-backend, but feels like it should be more central, as it will beused by the SPSH-backend-for-frontend and dBildungs-AIM. We should move it into shared or a seperate module

alweber-cap (approved):

src/modules/health/keycloak.health-indicator.ts
```diff
@@ -0,0 +1,28 @@
+import { HealthIndicator, HealthIndicatorResult } from '@nestjs/terminus';
+import { Injectable } from '@nestjs/common';
+import { LoginService } from '../ui-backend/domain/login.service.js';
+
+@Injectable()
+export class KeycloakHealthIndictor extends HealthIndicator {
```
WahlMartin, line 6:
```suggestion
export class KeycloakHealthIndicator extends HealthIndicator {
```

WahlMartin (changes requested):

PR #96, opened 2023-11-14
SPSH-144 Add hierarchical structure to organisation

marode-cap:
# Description
Adds two parents to organizations (administration and owner) to hierarchically structure organizations.
Additional endpoints to query by these parents / update relationships were also added.

Don't forget to initialize the DB to update the table structure.

The root organization is configured in `config.dev.json -> DATA.ROOT_ORGANISATION_ID`. Running `npm run setup` will create this organisation by default

- https://ticketsystem.dbildungscloud.de/browse/SPSH-144

## Approval for review
- [x] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.


src/modules/organisation/domain/organisation.service.ts
```diff
@@ -49,4 +70,96 @@ export class OrganisationService {
             items: organisations,
         };
     }
+
+    public async setAdministriertVon(parentId: string, childId: string): Promise<Result<void, DomainError>> {
+        const parentOrganisation: Option<OrganisationDo<true>> = await this.organisationRepo.findById(parentId);
```
kristoff-kiefer, line 75: I think it advisable to introduce a exists(id) call to the repo and use that.
Just finding something, mapping it and then discarding just to assert its existence seems wasteful to me.

Also when introducing such a call I believe we could leverage similar simplifications on a repo level
Please ponder on this and investigate feasability

test/config.test.json
```diff
@@ -32,6 +32,9 @@
         "PASSWORD": "password",
         "USE_TLS": false
     },
+    "DATA": {
+        "ROOT_ORGANISATION_ID": "id-of-the-root-organisation"
```
kristoff-kiefer, line 36: This will break as soon as we introduce a check for UUID. Please generate a random one (websites for that) and use it

kristoff-kiefer (changes requested):

kristoff-kiefer (approved):

PR #146, opened 2023-12-05
Spsh 227 fix kc deployment

kristoff-kiefer:
# Description
Remove KC from direct dependencies of server
Add labels for app name and version
make servername a mandatory parameter on deployment
Change KC-Config so that the client is set up properly
## Approval for review


charts/dbildungs-iam-keycloak-dev/dev-realm-spsh.json
DPDS93CT, line 1: so for my understanding, with publicClient:false and bearerOnly:false we're talking about confidential:true correct?

pkleybolte (approved):

pkleybolte (approved):

PR #347, opened 2024-03-05
Dbp 647 Rename all dbildungs-iam to dbildungs-iam-server and rename secret

aimee-889:
# Description
- rename various files and references to unify references
- add a GHA to be able to release a Image and Helm Chart as a Release based on a Tag 

## Approval for review


charts/dbildungs-iam-server/values.yaml
```diff
@@ -1,18 +1,16 @@
-dbildungsIamExternalPort: 80
-dbildungsIamCpuMax: 2
-dbildungsIamMemMax: 4G
-dbildungsIamRedisCpuMax: 2
-dbildungsIamRedisMemMax: 4G
-dbildungsIamRedisCpuRequest: 200m
-dbildungsIamRedisMemRequest: 200Mi
-dbildungsIamReplications: 1
+dbildungsIamServerCpuMax: 2
```
dimapin, line 1: Please define resource requests and limits with usual pattern:
resources: 
  requests:
  limits:

dimapin (changes requested):

charts/dbildungs-iam-server/values.yaml
aimee-889: If possible I would only like to rename them here to get rid of the old expression. There are many things that need to get adjusted in this Helm Chart, I would like to do it in a follow up ticket. The primary focus on these PRs is to get the naming itself straight.

dimapin (approved):

dbildungs-iam-server-secret.template.yaml
```diff
@@ -5,7 +5,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: dbildungs-iam-secrets
+  name: dbildungs-iam-server
```
M-Schiborr, line 8: I think this file should get deleted, it is not up to date and not needed anymore, because there is a secret definition file in the helm chart.

M-Schiborr (approved):

dbildungs-iam-server-secret.template.yaml
aimee-889: Thanks! Deleted it :)

PR #466, opened 2024-04-25
SPSH-644-suchstring-optional

YoussefBouch:
# Description
Search strings are now optional but can also handle an empty string.

https://ticketsystem.dbildungscloud.de/browse/SPSH-644

## Approval for review

### DIFF
--- a/src/modules/personenkontext/api/dbiam-personenkontext-filter.controller.ts
+++ b/src/modules/personenkontext/api/dbiam-personenkontext-filter.controller.ts
@@ -62,9 +62,10 @@ export class DbiamPersonenkontextFilterController {
         @Query() params: FindPersonenkontextSchulstrukturknotenBodyParams,
     ): Promise<FindSchulstrukturknotenResponse> {
         const anlage: PersonenkontextAnlage = this.personenkontextAnlageFactory.createNew();
+        const sskName: string = params.sskName || '';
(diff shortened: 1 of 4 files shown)


src/modules/personenkontext/api/dbiam-personenkontext-filter.controller.ts
```diff
@@ -62,9 +62,10 @@ export class DbiamPersonenkontextFilterController {
         @Query() params: FindPersonenkontextSchulstrukturknotenBodyParams,
     ): Promise<FindSchulstrukturknotenResponse> {
         const anlage: PersonenkontextAnlage = this.personenkontextAnlageFactory.createNew();
+        const sskName: string = params.sskName || '';
```
marode-cap, line 65: Use `??` instead to make it more clear (Won't make a difference in this case, but an empty string will also resolve to false when testing with `||`. For example, `"" || "test"` will resolve to `test`. The null-coalescing-operator is more idiomatic)

marode-cap (approved):

src/modules/personenkontext/api/dbiam-personenkontext-filter.controller.ts
YoussefBouch: Done.

PR #473, opened 2024-04-29
SPSH-605: Personen-Endpunkte absichern

DPDS93CT:
# Description
- use PersonPermissions on PersonController, to evaluate access on person related data for the logged-in/requesting user.

- https://github.com/dBildungsplattform/dbildungs-iam-server/pulls
- https://ticketsystem.dbildungscloud.de/browse/SPSH-605

## Approval for review


src/modules/person/domain/person.uc.ts
```diff
@@ -0,0 +1,49 @@
...
+        }
+
+        // Find all Personen on child-orgas (+root orgas)
+        const scope: PersonScope = new PersonScope()
+            .findBy({ organisationen: organisationIDs })
+            .sortBy('vorname', ScopeOrder.ASC);
+
+        const [persons]: Counted<Person<true>> = await this.personRepository.findBy(scope);
+
+        const person: Person<true> | undefined = persons.find((p: Person<true>) => p.id === personId);
```
marode-cap, line 43: Still not the biggest fan of this, because we always need to fetch EVERY person.
If we want to reuse the logic of the PersonScope, maybe expand the filters of `PersonScope.findBy` to accept a PersonID as well? Then the DB can handle finding the person and there should always only be 0 or 1 results and we don't need to find the person in the array.

src/modules/person/api/person.controller.ts
```diff
@@ -52,28 +52,74 @@ import { PersonendatensatzResponse } from './personendatensatz.response.js';
...
+        // Find all organisations where user has permission
+        const organisationIDs: OrganisationID[] | undefined = await permissions.getOrgIdsWithSystemrecht(
+            [RollenSystemRecht.PERSONEN_VERWALTEN],
+            true,
+        );
+        if (!organisationIDs || organisationIDs.length < 1) {
+            throw SchulConnexErrorMapper.mapSchulConnexErrorToHttpException(
+                SchulConnexErrorMapper.mapDomainErrorToSchulConnexError(new EntityNotFoundError('Person')),
+            );
+        }
```
marode-cap, line 122: `getOrgIdsWithSystemrecht` will never return `undefined`, can remove the extraneous checks here

src/modules/person/api/person.controller.ts
```diff
@@ -52,28 +52,74 @@ import { PersonendatensatzResponse } from './personendatensatz.response.js';
...
         @Inject(getMapperToken()) private readonly mapper: Mapper,
-    ) {}
+        config: ConfigService<ServerConfig>,
+    ) {
+        this.ROOT_ORGANISATION_ID = config.getOrThrow<DataConfig>('DATA').ROOT_ORGANISATION_ID;
+    }
+
+    //check that logged-in user is allowed to update person
+    private async getPersonOrThrowException(
+        params: PersonByIdParams,
```
marode-cap, line 84: Any reason why this accepts the entire params-object instead of just a `PersonID`?

src/modules/person/api/person.controller.ts
```diff
@@ -110,10 +156,15 @@ export class PersonController {
...
     @ApiForbiddenResponse({ description: 'Insufficient permissions to get the person.' })
     @ApiInternalServerErrorResponse({ description: 'Internal server error while getting the person.' })
-    public async findPersonById(@Param() params: PersonByIdParams): Promise<PersonendatensatzResponse> {
+    public async findPersonById(
+        @Param() params: PersonByIdParams,
+        @Permissions() permissions: PersonPermissions,
+    ): Promise<PersonendatensatzResponse> {
+        //check that logged-in user is allowed to update person
+        await this.getPersonOrThrowException(params, permissions);
         const person: Option<Person<true>> = await this.personRepository.findById(params.personId);
```
marode-cap, line 168: Doesn't the `getPersonOrThrowException`-method already return the person? Why fetch it again using the repository?

src/modules/person/api/person.controller.ts
```diff
@@ -233,13 +301,17 @@ export class PersonController {
...
     @ApiInternalServerErrorResponse({ description: 'An internal server error occurred.' })
     public async updatePerson(
         @Param() params: PersonByIdParams,
         @Body() body: UpdatePersonBodyParams,
+        @Permissions() permissions: PersonPermissions,
     ): Promise<PersonendatensatzResponse> {
+        //check that logged-in user is allowed to update person
+        await this.getPersonOrThrowException(params, permissions);
+
         const person: Option<Person<true>> = await this.personRepository.findById(params.personId);
```
marode-cap, line 315: Same as above

src/modules/person/api/person.controller.ts
```diff
@@ -282,10 +354,15 @@ export class PersonController {
...
     @ApiInternalServerErrorResponse({ description: 'Internal server error.' })
     @UseInterceptors(ResultInterceptor)
-    public async resetPasswordByPersonId(@Param() params: PersonByIdParams): Promise<Result<string>> {
+    public async resetPasswordByPersonId(
+        @Param() params: PersonByIdParams,
+        @Permissions() permissions: PersonPermissions,
+    ): Promise<Result<string>> {
+        //check that logged-in user is allowed to update person
+        await this.getPersonOrThrowException(params, permissions);
         const person: Option<Person<true>> = await this.personRepository.findById(params.personId);
```
marode-cap, line 366: Same as above

src/modules/person/api/person.controller.spec.ts
```diff
@@ -30,13 +30,21 @@ import { Person } from '../domain/person.js';
...
 
 describe('PersonController', () => {
     let module: TestingModule;
     let personController: PersonController;
     let personenkontextUcMock: DeepMocked<PersonenkontextUc>;
     let personRepositoryMock: DeepMocked<PersonRepository>;
     let usernameGeneratorService: DeepMocked<UsernameGeneratorService>;
+    let personUcMock: DeepMocked<PersonUc>;
+    let personPermissionsMock: DeepMocked<PersonPermissions>;
+    //let configServiceMock: DeepMocked<ConfigService>;
```
marode-cap, line 47: Remove comment

src/modules/person/api/person.controller.spec.ts
```diff
@@ -65,14 +73,39 @@ describe('PersonController', () => {
...
+                    useValue: createMock<ConfigService>(),
+                },
             ],
         }).compile();
         personController = module.get(PersonController);
         personenkontextUcMock = module.get(PersonenkontextUc);
         personRepositoryMock = module.get(PersonRepository);
         usernameGeneratorService = module.get(UsernameGeneratorService);
+        personUcMock = module.get(PersonUc);
+        //configServiceMock = module.get(ConfigService);
```
marode-cap, line 91: Remove comment

src/modules/person/api/person.controller.spec.ts
```diff
@@ -65,14 +73,39 @@ describe('PersonController', () => {
...
+        personPermissionsMock = createMock<PersonPermissions>();
```
marode-cap, line 92: I wouldn't create this mock for all tests. Instead create the mock in every test if it is needed. This is basically global state without the need to be

src/modules/person/domain/person.uc.ts
```diff
@@ -0,0 +1,49 @@
...
+
+        if (!person) return { ok: false, error: new EntityNotFoundError('Person') };
+
+        return { ok: true, value: person };
+    }
```
marode-cap, line 48: If we already have this "helper"-function that checks our permissions for single persons, maybe it would be nice to also add a function for finding the persons? So everything is on one place, we will have the same pattern with other entities as well.

I'm thinking about a function that returns a filtered PersonScope, so the controller can just get that scope and use it.
Something like
```typescript
public async getPersonScopeWithPermissions(permissions: PersonPermissions): Promise<PersonScope> {
    let organisationIDs: OrganisationID[] | undefined = await permissions.getOrgIdsWithSystemrecht(
        [RollenSystemRecht.PERSONEN_VERWALTEN],
        true,
    );

    // Check if user has permission on root organisation
    if (organisationIDs?.includes(this.ROOT_ORGANISATION_ID)) {
        organisationIDs = undefined;
    }
    
    return new PersonScope()
        .findBy({ organisationen: organisationIDs })
        .setScopeWhereOperator(ScopeOperator.AND);
}
```

This way the controller can call `getPersonScopeWithPermissions` to get an already filtered scope and just ignore permissions. Other filters could be tacked on using `.findBy`, because they don't overwrite other filters. Would also remove the ROOT_ORGANISATION_ID from the controller.

Just a suggestion, this is not blocking my approval for this PR

marode-cap (changes requested):

src/modules/person/api/person.controller.ts
```diff
@@ -132,16 +162,21 @@ export class PersonController {
...
         @Body() bodyParams: CreatePersonenkontextBodyParams,
+        @Permissions() permissions: PersonPermissions,
     ): Promise<PersonenkontextResponse> {
         const personenkontextDto: CreatePersonenkontextDto = this.mapper.map(
             bodyParams,
             CreatePersonenkontextBodyParams,
             CreatePersonenkontextDto,
         );
+        //check that logged-in user is allowed to update person
+        await this.personUc.getPersonIfAllowed(pathParams.personId, permissions);
```
marode-cap, line 178: `personUc.getPersonIfAllowed` just returns an error if the user is not allowed to access the person, your discarding the result here

src/modules/person/persistence/person.scope.ts
```diff
@@ -4,6 +4,7 @@ import { PersonEntity } from './person.entity.js';
 import { OrganisationID } from '../../../shared/types/aggregate-ids.types.js';
 
 type FindProps = {
+    id: string;
```
marode-cap, line 7: Could use `PersonID` instead of `string`

marode-cap (approved):

PR #502, opened 2024-05-17
Spsh 666 benutzer loeschen

YoussefBouch:
# Description

## Approval for review

### DIFF
--- a/src/modules/person/api/person.controller.ts
+++ b/src/modules/person/api/person.controller.ts
@@ -3,6 +3,7 @@ import { getMapperToken } from '@automapper/nestjs';
 import {
     Body,
     Controller,
+    Delete,
     Get,
     HttpCode,
     HttpStatus,
@@ -22,6 +23,7 @@ import {
     ApiCreatedResponse,
     ApiForbiddenResponse,
     ApiInternalServerErrorResponse,
+    ApiNoContentResponse,
     ApiNotFoundResponse,
     ApiOAuth2,
     ApiOkResponse,
@@ -131,6 +133,33 @@ export class PersonController {
         }
 
         return new PersonendatensatzResponse(result, true);
+    }
+
+    @Delete(':personId')
+    @HttpCode(HttpStatus.NO_CONTENT)
+    @ApiNoContentResponse({
+        description: 'The person and all their kontexte were successfully deleted.',
+    })
+    @ApiBadRequestResponse({ description: 'Request has wrong format.' })
+    @ApiUnauthorizedResponse({ description: 'Request is not authorized.' })
+    @ApiNotFoundResponse({ description: 'The person was not found.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permissions to perform operation.' })
+    @ApiInternalServerErrorResponse({ description: 'An internal server error occurred.' })
+    public async deletePersonById(
+        @Param() params: PersonByIdParams,
+        @Permissions() permissions: PersonPermissions,
+    ): Promise<void> {
+
+        // First delete all kontexte for the personId
+        await this.personenkontextUc.deletePersonenkontexteByPersonId(params.personId);
+        // Delete the person after all their kontexte were deleted
+        const response: Result<void, DomainError> = await this.personUc.deletePersonIfAllowed(
+            params.personId,
+            permissions,
+        );
+        if (response instanceof SchulConnexError) {
+            throw SchulConnexErrorMapper.mapSchulConnexErrorToHttpException(response);
+        }
     }
 
     @Get(':personId')
--- a/src/modules/person/persistence/person.repository.ts
+++ b/src/modules/person/persistence/person.repository.ts
@@ -118,6 +118,12 @@ export class PersonRepository {
         return mapEntityToAggregateInplace(personEntity, personWithKeycloakUser);
     }
 
+    public async delete(person: Person<true>): Promise<number> {
+        const deletedPersons: number = await this.em.nativeDelete(PersonEntity, person.id);
+
+        return deletedPersons;
+    }
+
     public async update(person: Person<true>): Promise<Person<true> | DomainError> {
         const personEntity: Loaded<PersonEntity> = await this.em.findOneOrFail(PersonEntity, person.id);
 
--- a/src/modules/personenkontext/domain/personenkontext.service.ts
+++ b/src/modules/personenkontext/domain/personenkontext.service.ts
@@ -133,4 +133,16 @@ export class PersonenkontextService {
 
         return { ok: true, value: undefined };
     }
+
+    public async deletePersonenkontexteByPersonId(personId: string): Promise<Result<void, DomainError>> {
+
+        const deletedPersonenkontexte: number = await this.dBiamPersonenkontextRepo.deleteByPerson(personId);
+
+        if (deletedPersonenkontexte === 0) {
+            return { ok: false, error: new EntityCouldNotBeDeleted('Personenkontext', personId) };
+        }
+
+        return { ok: true, value: undefined };
+    }
+
 }
--- a/src/modules/personenkontext/domain/personenkontext.service.ts
+++ b/src/modules/personenkontext/domain/personenkontext.service.ts
@@ -135,14 +135,16 @@ export class PersonenkontextService {
     }
 
     public async deletePersonenkontexteByPersonId(personId: string): Promise<Result<void, DomainError>> {
+        const personenkontexte: Personenkontext<true>[] = await this.dBiamPersonenkontextRepo.findByPerson(personId);
 
         const deletedPersonenkontexte: number = await this.dBiamPersonenkontextRepo.deleteByPerson(personId);
 
-        if (deletedPersonenkontexte === 0) {
-            return { ok: false, error: new EntityCouldNotBeDeleted('Personenkontext', personId) };
+        if (deletedPersonenkontexte !== personenkontexte.length) {
+            // Create a list of the kontexte IDs that were not deleted
+            const notDeletedEntities: string[] = personenkontexte.map((entity: Personenkontext<true>) => entity.id);
+            return { ok: false, error: new EntityCouldNotBeDeleted('Personenkontext', notDeletedEntities.join(', ')) };
         }
 
         return { ok: true, value: undefined };
     }
-
 }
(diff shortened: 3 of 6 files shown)


src/modules/personenkontext/persistence/dbiam-personenkontext.repo.ts
```diff
@@ -43,6 +43,14 @@ export class DBiamPersonenkontextRepo {
         return personenKontexte.map(mapEntityToAggregate);
     }
 
+    public async deleteByPerson(personId: PersonID): Promise<number> {
```
DPDS93CT, line 46: This method seems not to be tested.

src/modules/personenkontext/domain/personenkontext.service.ts
```diff
@@ -133,4 +133,18 @@ export class PersonenkontextService {
 
         return { ok: true, value: undefined };
     }
+
+    public async deletePersonenkontexteByPersonId(personId: string): Promise<Result<void, DomainError>> {
```
DPDS93CT, line 137: This method seems not to be tested.

src/modules/person/api/person.controller.ts
DPDS93CT, line 1: Isn't there too much going on in the new method (deletePersonById) for a Controller-method?

Also it's seems not be covered fully by tests

src/shared/error/person-has-no-keycloak-id.ts
DPDS93CT, line 1: We don't want such files in shared, maybe the keycloak module is a better location for this.

Also suggest renaming to ...-id.error.ts

src/providers/keycloak-user-provider.ts
DPDS93CT, line 1: Should this explicitly be located in /src/provider or can it be places in src/keycloka/provider or src/keycloak/event ?

(The LDAP event handling stuff will be located in the corresponding ldap-directory src/modules/ldap/...)

DPDS93CT (changes requested):

src/modules/person/api/person.controller.ts
YoussefBouch: Imho it does what it should, and that is publishing the event, calling the service to delete the contexts and then the other service to delete the person (In case of errors then it throws an error). That's it. Otherwise I don't see how to make it shorter without adding unncessesary methods somewhere else.

src/modules/person/persistence/person.repository.ts
```diff
@@ -74,10 +88,31 @@ export function mapEntityToAggregateInplace(entity: PersonEntity, person: Person
...
         private readonly kcUserService: KeycloakUserService,
         private readonly em: EntityManager,
-    ) {}
+        config: ConfigService<ServerConfig>,
+        private readonly dBiamPersonenkontextRepo: DBiamPersonenkontextRepo,
+    ) {
+        this.ROOT_ORGANISATION_ID = config.getOrThrow<DataConfig>('DATA').ROOT_ORGANISATION_ID;
+    }
+
+    public async getPersonScopeWithPermissions(permissions: PersonPermissions): Promise<PersonScope> {
```
pkleybolte, line 102: Please change to `private`. The scope is tightly coupled to mikro-orm and should not be accessible outside of the repo.

src/modules/person/persistence/person.repository.ts
```diff
@@ -118,6 +185,12 @@ export class PersonRepository {
         return mapEntityToAggregateInplace(personEntity, personWithKeycloakUser);
     }
 
+    public async delete(person: Person<true>): Promise<number> {
```
pkleybolte, line 188: can this be removed or at least set private? It should not be possible to delete without permissions.

src/modules/person/persistence/person.repository.ts
```diff
@@ -170,4 +243,32 @@ export class PersonRepository {
...
+    ): Promise<Result<void, DomainError>> {
+        // Check if the person has a keycloakUserId
+        if (!person.keycloakUserId) {
+            throw new PersonHasNoKeycloakId(person.id);
+        }
+        // Delete the person from Keyclock
+        await this.kcUserService.delete(person.keycloakUserId);
+        // First, delete all kontexte for the personId
+        const kontextResponse: Result<void, DomainError> =
+            await this.dBiamPersonenkontextRepo.deletePersonenkontexteByPersonId(person.id);
```
pkleybolte, line 260: Like this there is no transaction around the deletion of a person and her kontexts. That is not nice, but acceptable.
In this case it is fine to set `cascade: [Cascade.REMOVE]` in the person.entity.ts in the personenkontext's `@OneToMany`

src/modules/personenkontext/persistence/dbiam-personenkontext.repo.ts
```diff
@@ -43,6 +45,14 @@ export class DBiamPersonenkontextRepo {
...
```
pkleybolte, line 48: Not needed if cascade is used. Otherwise permissions should be checked.

src/modules/personenkontext/persistence/dbiam-personenkontext.repo.ts
```diff
@@ -113,4 +123,18 @@ export class DBiamPersonenkontextRepo {
 
         return mapEntityToAggregate(personenKontextEntity);
     }
+
+    public async deletePersonenkontexteByPersonId(personId: string): Promise<Result<void, DomainError>> {
+        const personenkontexte: Personenkontext<true>[] = await this.findByPerson(personId);
```
pkleybolte, line 128: Not needed if cascade is used. Otherwise permissions should be checked.

pkleybolte (changes requested):

src/modules/person/persistence/person.repository.ts
YoussefBouch: Does cascade: [Cascade.REMOVE] really helps with anything in our case? The personId is a foreign key in the entity personkontexte that references the persons entity "id" so you can't really delete a person if it has at least 1 kontext still available.

src/modules/person/person-api.module.ts
```diff
@@ -24,6 +25,7 @@ import { PersonApiMapper } from './mapper/person-api.mapper.js';
         PersonenKontextModule,
         OrganisationModule,
         KeycloakAdministrationModule,
+        EventModule,
```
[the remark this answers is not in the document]
YoussefBouch: It isn't

src/shared/error/index.ts
```diff
@@ -12,3 +12,4 @@ export * from './entity-already-exists.error.js';
 export * from './invalid-name.error.js';
 export * from './invalid-character-set.error.js';
 export * from './invalid-attribute-length.error.js';
+export * from '../../modules/keycloak-administration/error/person-has-no-keycloak-id.error.js';
```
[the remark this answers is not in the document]
YoussefBouch: Right.

src/modules/person/persistence/person.repository.ts
YoussefBouch: Done.

src/modules/personenkontext/persistence/dbiam-personenkontext.repo.ts
YoussefBouch: Permissions to retrieve the kontexte is not in scope of this ticket I think? We implemented /personen endpoints under SPSH-605. Maybe we should do the same and secure the /personenkontexte.

src/modules/person/persistence/person.repository.ts
YoussefBouch: Removed, not necessary.

src/modules/personenkontext/persistence/dbiam-personenkontext.repo.ts
YoussefBouch: Ps: To check for permissions for kontexte there is a Ticket still in progress, so we can't do that here.

src/modules/person/persistence/person.repository.ts
```diff
@@ -170,4 +234,25 @@ export class PersonRepository {
 
         return person;
     }
+
+    public async deletePersonAndKontexte(
```
pkleybolte, line 238: Why don't you use deletePersonIfAllowed directly? Especially in this case it is necessary, because you can read the person with "Person verwalten" right, but you can delete it only with "Person sofort löschen".

src/modules/person/persistence/person.repository.ts
```diff
@@ -95,6 +127,38 @@ export class PersonRepository {
...
+
+        const [persons]: Counted<Person<true>> = await this.findBy(scope);
+        const person: Person<true> | undefined = persons[0];
+
+        if (!person) return { ok: false, error: new EntityNotFoundError('Person') };
+
+        return { ok: true, value: person };
+    }
+
+    public async deletePersonIfAllowed(
```
pkleybolte, line 142: "IfAllowed" should not be part of the name, because every repo method should only work if the user is allowed.

pkleybolte (changes requested):

src/modules/personenkontext/persistence/personenkontext.entity.ts
```diff
@@ -8,7 +8,13 @@ import { PersonEntity } from '../../person/persistence/person.entity.js';
 @Unique({ properties: ['personId', 'organisationId', 'rolleId'] })
 export class PersonenkontextEntity extends TimestampedEntity {
     @AutoMap()
-    @ManyToOne({ columnType: 'uuid', cascade: [], ref: true, nullable: false, entity: () => PersonEntity })
+    @ManyToOne({
+        columnType: 'uuid',
+        cascade: [Cascade.REMOVE],
```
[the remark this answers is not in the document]
YoussefBouch: It's necessary otherwise cascading doesn't work. (tried it in 1 direction). I think it's because it's a bidirecitonal relationship?

Also if we delete the Personenkontext the person isn't deleted since the person is the parent.

src/modules/person/persistence/person.repository.ts
YoussefBouch: Because the method checks for permission to read the user before deletion so I thought it made sense. But you are right, maybe it's 'redundant' to mention it again.

src/modules/person/persistence/person.repository.ts
YoussefBouch: Done

pkleybolte (approved):

DPDS93CT (approved):

PR #535, opened 2024-06-05
SPSH-624 Implemented itslearning client

marode-cap:
# Description
Implements a client to interact with the ItsLearning IMS ES SOAP API

## Approval for review


charts/dbildungs-iam-server/config/config.json
```diff
@@ -45,5 +45,11 @@
     "KEYCLOAK_ADMINISTRATION_MODULE_LOG_LEVEL": "debug",
     "HEALTH_MODULE_LOG_LEVEL": "debug",
     "BACKEND_FOR_FRONTEND_MODULE_LOG_LEVEL": "debug"
+  },
+  "ITSLEARNING": {
+    "ENABLED": false,
+    "ENDPOINT": "https://localhost",
+    "USERNAME": "username",
```
kristoff-kiefer, line 52: hmmm, looks a little weird as a default, maybe https://itslearning.example.com would be better to show that it is indeed **not** a local service

config/config.json
```diff
@@ -53,5 +53,11 @@
         "KEYCLOAK_ADMINISTRATION_MODULE_LOG_LEVEL": "debug",
         "HEALTH_MODULE_LOG_LEVEL": "debug",
         "BACKEND_FOR_FRONTEND_MODULE_LOG_LEVEL": "debug"
+    },
+    "ITSLEARNING": {
+        "ENABLED": false,
+        "ENDPOINT": "https://localhost",
```
kristoff-kiefer, line 59: Similar to above, the Demo-URL could and should be different maybe?

src/modules/itslearning/actions/base-action.ts
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+        Body: BodyResponse;
+    };
+};
+
+export abstract class IMSESAction<ResponseBodyType, ResultType> {
+    protected readonly xmlBuilder: XMLBuilder = new XMLBuilder({ ignoreAttributes: false });
+
+    protected readonly xmlParser: XMLParser = new XMLParser({
```
kristoff-kiefer, line 30: Think about injecting that maybe?

src/modules/itslearning/actions/read-all-persons.action.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+                pageIndex: 1,
+                pageSize: 10,
+            });
+
+            expect(action.buildRequest()).toBeDefined();
+        });
+    });
+
+    describe('isArrayOverride', () => {
+        it('should return true for specific tags', () => {
```
kristoff-kiefer, line 17: be more specific about "specific"

kristoff-kiefer (changes requested):

kristoff-kiefer (approved):

src/modules/itslearning/actions/base-action.ts
marode-cap: The parser may need specific configuration per action (currently only the override for 'isArray', which tells the parser if a tag should always be treated as a list, even if only one entry is present)

Injecting an already constructed parser would also mean you can't easily construct instances of the different actions

PR #553, opened 2024-06-20
Spsh 696: Rolle bearbeiten

phaelcg:
# Description

Implemented a new endpoint: PUT/api/rolle/{rolleId}
Implemented all the checks for the update in the Rolle aggregate and rolleRepo.
Implemented the unit tests.

https://ticketsystem.dbildungscloud.de/browse/SPSH-696

## Approval for review


src/modules/rolle/repo/rolle.repo.ts
```diff
@@ -216,6 +216,39 @@ export class RolleRepo {
         }
     }
 
+    public async updateRolle(
```
kristoff-kiefer, line 219: Can you formulate the checks made as a spec and just slap it into update?

Also: Is it possible to include update in the save method? The repo should be able to figure out whether to do update or save on its own. (Based on if we give it an ID or no

kristoff-kiefer (changes requested):

src/modules/rolle/repo/rolle.repo.ts
phaelcg: Which checks? If you mean the check 'isRolleAlreadyAssigned' which is in the controller, we cannot move it in the RolleRepo because it creates a circular reference error because of the DBiamPersonenkontextRepo.

We have already a save method that does call create or save based on the ID of the aggregate. This new method updateRolle can be maybe renamed to 'updateRolleAuthorized' because it also checks the permissions & other things.

kristoff-kiefer (approved):

PR #565, opened 2024-06-28
SPSH-387:  Anmeldung von Benutzern ablehnen, die uns nicht bekannt sind

phaelcg:
… Keycloak User is not found in the DB with help of custom errors and also unit tests.

# Description
Implemented the check for the existence of keycloak users for the OpenIdConnectStrategy
Implemented the redirect to a FE custom error page when the Keycloak User is not found  in the DB in the LoginGuard 
Implemented custom errors and unit tests.

https://ticketsystem.dbildungscloud.de/browse/SPSH-387

## Approval for review


src/modules/authentication/passport/oidc.strategy.ts
```diff
@@ -40,7 +50,11 @@ export class OpenIdConnectStrategy extends PassportStrategy(Strategy, 'oidc') {
             };
             return user;
         } catch (err: unknown) {
-            throw new UnauthorizedException();
+            if (err instanceof KeycloakUserNotFoundError) {
+                throw err;
+            } else {
+                throw new UnauthorizedException();
+            }
```
marode-cap, line 57: Rethrowing errors is not pretty. Maybe only include the `client.userinfo` and `personRepo.findByKeycloakUserId` inside the try-catch and move the does-this-person-exist check outside.

phaelcg: I agree that rethrowing the exception is not pretty but if this.personRepo.findByKeycloakUserId() throws an Exception it has to be mapped into an UnauthorizedException() otherwise an unknown Exception will be bubled up to the FE.

marode-cap (approved):

PR #572, opened 2024-07-03
Spsh 543: Client für Swagger mit längeren Timeouts

phaelcg:
# Description
Added a test client for keycloak in the dev-realm-spsh.json.
Configured swagger to use the test client-id in the dev/test environment.
Updated the unit-tests for the Keycloak Configuration.

https://ticketsystem.dbildungscloud.de/browse/SPSH-543

## Datasecurity
The test client should only be created in keycloak in the dev/test environment

## Approval for review


config/dev-realm-spsh.json
marode-cap, line 1: This file is only used for local environments I think? Needs to be included here as well:
https://github.com/dBildungsplattform/dbildungs-iam-keycloak/blob/main/charts/dbildungs-iam-keycloak/dev-realm-spsh.json

But this needs careful consideration, otherwise the new client will also be available in staging and production

config/dev-realm-spsh.json
```diff
@@ -1005,7 +1006,116 @@
...
+        "adminUrl": "",
+        "baseUrl": "https://localhost:8099/",
+        "surrogateAuthRequired": false,
+        "enabled": true,
+        "alwaysDisplayInConsole": false,
+        "clientAuthenticatorType": "client-secret",
+        "secret": "YDp6fYkbUcj4ZkyAOnbAHGQ9O72htc5M",
+        "redirectUris": [
+           "https://127.0,0.1:8099/*",
+			"http://127.0,0.1:9090/*",
```
marode-cap, line 1025: Second URI has a comma in it

src/server/main.ts
```diff
@@ -56,7 +56,10 @@ async function bootstrap(): Promise<void> {
         swaggerOptions: {
             persistAuthorization: false,
             initOAuth: {
-                clientId: keycloakConfig.CLIENT_ID,
+                clientId:
+                    process.env['DEPLOY_STAGE'] === 'dev' || process.env['NODE_ENV'] === 'dev'
+                        ? keycloakConfig.TEST_CLIENT_ID
+                        : keycloakConfig.CLIENT_ID,
```
marode-cap, line 62: On the cluster both of these environment variables are set to "prod", even for our test-branches (Like the one for this PR).

marode-cap (changes requested):
This is supposed to help testers on the deployed branches, please make sure it works there.

config/dev-realm-spsh.json
phaelcg: Exactly, the client cannot be deployed to stage or prod, that's why it needs to be clarified with dev-ops

marode-cap (approved):

PR #600, opened 2024-07-30
Spsh 308

AlexanderUngefug:


src/modules/authentication/api/authentication.controller.ts
```diff
@@ -118,4 +122,21 @@ export class AuthenticationController {
         );
         return new UserinfoResponse(permissions, rolleFieldsResponse);
     }
+
+    @Get('resetPassword')
```
psachmann, line 126: Better call the endpoint **reset-password**, it's better readable and prevents issues with case sensitivity or insensitivity.

src/modules/authentication/api/authentication.controller.ts
```diff
@@ -118,4 +122,21 @@ export class AuthenticationController {
...
+        const responseType: string = 'code';
+        const scope: string = 'openid';
+        const kcAction: string = 'UPDATE_PASSWORD';
+        const setNewPasswordUrl: string = `${this.client.issuer.metadata.authorization_endpoint}?client_id=${clientId}&response_type=${responseType}&scope=${scope}&kc_action=${kcAction}&redirect_uri=${redirectUrl}`;
+        res.redirect(setNewPasswordUrl);
+    }
+
+    public getKeyCloakclientRealm(): string {
+        return this.keyCloakclientRealm;
+    }
```
psachmann, line 141: Is this getter needed? You can also write a native getter like this:
```typescript
public get keycloakClientRealm(): string {
    return this.keyCloakclientRealm;
}
```

psachmann (changes requested):

psachmann (approved):

PR #610, opened 2024-08-09
SPSH-830 Update Klassen in itslearning

marode-cap:
# Description
Add event handlers to for updates and deletes of klassen.

## Approval for review


src/modules/itslearning/actions/create-memberships.action.ts
```diff
@@ -0,0 +1,66 @@
+import { DomainError } from '../../../shared/error/domain.error.js';
+import { IMS_COMMON_SCHEMA, IMS_MEMBER_MAN_DATA_SCHEMA, IMS_MEMBER_MAN_MESS_SCHEMA } from '../schemas.js';
+import { IMSESRoleType } from '../types/role.enum.js';
+import { IMSESMassAction } from './base-mass-action.js';
+
+// Incomplete
```
kristoff-kiefer, line 6: What do you mean by that?

kristoff-kiefer (changes requested):

src/modules/itslearning/actions/create-memberships.action.ts
marode-cap: The real action can take more inputs than this, but I've limited it to the inputs we need. 
Therefore, the properties shown here are incomplete.

I've used this convention in the other actions as well

pkleybolte (approved):

PR #613, opened 2024-08-13
SPSH-841

DeFi3298:
Implementation of Hardware-Token assignment for TwoFactorAuthentification (Backend Part)
Ticket: https://ticketsystem.dbildungscloud.de/browse/SPSH-841


src/modules/privacy-idea-administration/privacy-idea-administration.controller.ts
```diff
@@ -89,4 +92,57 @@ export class PrivacyIdeaAdministrationController {
...
+    @ApiUnauthorizedResponse({ description: 'Not authorized to assign hardware token.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permissions to reset token.' })
+    @ApiNotFoundResponse({ description: 'Insufficient permissions to assign hardware token.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while assigning a hardware token.' })
+    @Public()
+    public async assignHardwareToken(
+        @Body() params: AssignHardwareTokenBodyParams,
+    ): Promise<AssignHardwareTokenResponse | undefined> {
+        try {
+            const result: AssignTokenResponse = await this.privacyIdeaAdministrationService.assignHardwareToken(
```
marode-cap, line 112: Like in the other PR, this endpoint ignores the logged in user.

If this is not a part of this ticket, just resolve this comment

src/modules/privacy-idea-administration/privacy-idea-administration.controller.ts
```diff
@@ -89,4 +92,57 @@ export class PrivacyIdeaAdministrationController {
...
+                );
+            }
+            // Handle other unexpected errors
+            throw new HttpException(
+                {
+                    statusCode: HttpStatus.INTERNAL_SERVER_ERROR,
+                    message: 'An unexpected error occurred.',
+                },
+                HttpStatus.INTERNAL_SERVER_ERROR,
+            );
```
marode-cap, line 145: We try to use Controller-Specific error-filters so we don't have to do things like this. 
i.e. something like `PersonExceptionFilter`. Maybe add an ExceptionFilter to the entire PrivacyIdeaAdministrationController to handle errors that can occur in here.

src/modules/privacy-idea-administration/privacy-idea-administration.service.spec.ts
```diff
@@ -79,6 +86,168 @@ export const mockNonErrorThrow = (): never => {
     throw { message: mockErrorMsg };
 };
 
+export const mockTokenVerificationResponse: TokenVerificationResponse = {
```
marode-cap, line 89: Do these mock-data-objects need to be exported? 
Exporting things from test-files should not be done

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
```diff
@@ -198,4 +203,99 @@ export class PrivacyIdeaAdministrationService {
             }
         }
     }
+
+    private async verifyTokenStatus(serial: string, token: string): Promise<TokenVerificationResponse> {
+        const endpoint: string = `/token?serial=${serial}`;
+        const baseUrl: string = process.env['PI_BASE_URL'] ?? 'http://localhost:5000';
+        const url: string = `${baseUrl}${endpoint}`;
+        const headers: { Authorization: string } = {
```
marode-cap, line 211: Why is PI_BASE_URL taken from the environment-variables and not our ConfigService? 
Wouldn't need to have a default here as well, because the default can be in our well-structured config.json

Also, I'd recommend using the standard `URL` class to construct the url, instead of string concatenation

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
```diff
@@ -198,4 +203,99 @@ export class PrivacyIdeaAdministrationService {
...
+                'general-token-error',
+            );
+        }
+    }
+
+    private async getSerialWithOTP(otp: string, token: string): Promise<TokenOTPSerialResponse> {
+        const endpoint: string = `/token/getserial/${otp}?unassigned=1`;
+        const baseUrl: string = process.env['PI_BASE_URL'] ?? 'http://localhost:5000';
+        const url: string = `${baseUrl}${endpoint}`;
+        const headers: { Authorization: string } = {
```
marode-cap, line 231: See comment above

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
```diff
@@ -198,4 +203,99 @@ export class PrivacyIdeaAdministrationService {
...
+                'general-token-error',
+            );
+        }
+    }
+
+    private async assignToken(serial: string, token: string, user: string): Promise<AssignTokenResponse> {
+        const endpoint: string = '/token/assign';
+        const baseUrl: string = process.env['PI_BASE_URL'] ?? 'http://localhost:5000';
+        const url: string = baseUrl + endpoint;
+        const headers: { Authorization: string; 'Content-Type': string } = {
```
marode-cap, line 252: See comment above

marode-cap (changes requested):

src/modules/privacy-idea-administration/privacy-idea-administration.controller.ts
DeFi3298: check implemented

src/modules/privacy-idea-administration/privacy-idea-administration.controller.ts
DeFi3298: PrivacyIdeaAdministrationExceptionFilter implemented -> Frontend-Part work in progress

src/modules/privacy-idea-administration/privacy-idea-administration.service.spec.ts
DeFi3298: Removed

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
DeFi3298: Adjusted with Config

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
DeFi3298: Adjusted with Config

src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
DeFi3298: Adjusted with Config

marode-cap (approved):

PR #619, opened 2024-08-20
Spsh 919

clauyan:
# Description

Enable admins to lock and unlock users. The locking is triggered via the schulportal-UI, but actually happens in Keycloak. Information about who triggered the lock and when is stored in Keycloak via custom attributes.

https://ticketsystem.dbildungscloud.de/browse/SPSH-919
https://ticketsystem.dbildungscloud.de/browse/SPSH-920
https://github.com/dBildungsplattform/schulportal-client/pull/215

## Changes

- adds a new route to `PersonController `
- `KeycloakUserService` handles the communication with keycloak for locks
- `PersonRepository` contains a new function to extend a `Person` with information from Keycloak
- `OrganisationRepository` contains a new function to get the list of parent organisationen from a list of organisationen


src/modules/keycloak-administration/domain/keycloak-client/user-representation.dto.ts
```diff
@@ -15,4 +15,9 @@ export class UserRepresentationDto {
 
     @IsNumber()
     public createdTimestamp!: number;
+
+    @IsBoolean()
+    public enabled!: boolean;
+
+    public attributes!: Record<string, string>;
 }
```
marode-cap, line 23: If I recall correctly, the attributes returned by keycloak are actually `Record<string, string[]>`.
Could you double-check this?

src/modules/keycloak-administration/domain/keycloak-client/user.mapper.profile.ts
marode-cap, line 1: This file is empty

src/modules/keycloak-administration/domain/user.ts
```diff
@@ -25,7 +27,9 @@ export class User<WasPersisted extends boolean> {
         email: string | undefined,
         createdDate: Date,
         externalSystemIDs: ExternalSystemIDs,
+        enabled: boolean,
+        attributes: Record<string, string>,
     ): User<WasPersisted> {
-        return new User(id, username, email, createdDate, externalSystemIDs);
+        return new User(id, username, email, createdDate, externalSystemIDs, enabled, attributes);
```
marode-cap, line 33: externalSystemIDs also use the keycloak attributes. Just fyi, since we may need to separate these two.
Maybe it would be better to explicitly have a "disabledReason" and not the entire attributes?

src/modules/person/api/lock-user.body.params.ts
```diff
@@ -0,0 +1,18 @@
...
+        nullable: false,
+    })
+    public readonly lock!: boolean;
+
+    @IsString()
+    @ApiProperty({
+        required: true,
+        nullable: false,
+    })
+    public readonly locked_from!: string;
```
marode-cap, line 17: Maybe also include a "@IsNotEmpty()", just in case

src/modules/person/api/person.controller.ts
```diff
@@ -438,4 +449,55 @@ export class PersonController {
...
+            customAttributes,
+        );
+        if (!result.ok) {
+            throw SchulConnexErrorMapper.mapSchulConnexErrorToHttpException(
+                SchulConnexErrorMapper.mapDomainErrorToSchulConnexError(
+                    new EntityCouldNotBeUpdated('Person', personId),
+                ),
+            );
+        }
+        return new PersonLockResponse(`User has been successfully ${lockUserBodyParams.lock ? '' : 'un'}locked.`);
```
marode-cap, line 501: This is a new endpoint and we don't need to adhere to schulconnex spec (right?).
Prefer throwing our domain errors directly and handle them in the specific filters (We already have `AuthenticationExceptionFilter` and `PersonExceptionFilter` in this controller, maybe expand those if needed)

src/modules/person/api/person.controller.ts
```diff
@@ -438,4 +449,55 @@ export class PersonController {
...
+
+        if (!personResult.value?.keycloakUserId) {
+            throw SchulConnexErrorMapper.mapSchulConnexErrorToHttpException(
+                SchulConnexErrorMapper.mapDomainErrorToSchulConnexError(new EntityNotFoundError('Person', personId)),
+            );
+        }
+
+        const tempAttributes: Record<string, string> = {};
+        tempAttributes[LOCK_KEYS[0] as string] = lockUserBodyParams.locked_from;
+        tempAttributes[LOCK_KEYS[1] as string] = new Date().toISOString();
```
marode-cap, line 485: `as` casts are not great. I'd recommend defining `LOCK_KEYS` as a tuple (`[string, string]`) then TS will know indices 0 and 1 are valid.

src/modules/person/persistence/person.repository.ts
```diff
@@ -148,13 +150,28 @@ export class PersonRepository {
...
+        if (!person.keycloakUserId) {
+            return person;
+        }
+
+        const keyCloakUserDataResponse: Result<User<true>, DomainError> = await this.kcUserService.findById(
+            person.keycloakUserId,
+        );
+        if (keyCloakUserDataResponse.ok) {
+            person.attributes = keyCloakUserDataResponse.value.attributes;
+            person.isLocked = keyCloakUserDataResponse.value.enabled === false;
```
marode-cap, line 170: Like I've said in another comment, `attributes` might contain additional Data for external systems. Is probably not a problem here, but something you should be aware of

marode-cap (changes requested):

src/modules/keycloak-administration/domain/keycloak-client/user.mapper.profile.ts
clauyan: Deleted

src/modules/keycloak-administration/domain/user.ts
clauyan: I added a more specific type here.

src/modules/person/api/lock-user.body.params.ts
clauyan: Done!

src/modules/person/persistence/person.repository.ts
clauyan: Should not be a problem anymore, because the lock-attributes are a separate type now.

src/modules/person/api/person.controller.ts
clauyan: I refactored this to throw domain errors, two of which are new. @marode-cap could you please verify if I implemented this correctly?

marode-cap: @clauyan looks okay from my perspective

marode-cap (approved):

PR #641, opened 2024-09-05
SPSH-986

DPDS93CT:
# Description

## Approval for review


migrations/Migration20240909061707.ts
```diff
@@ -0,0 +1,20 @@
+import { Migration } from '@mikro-orm/migrations';
+
+export class Migration20240909061707 extends Migration {
+
+  public up(): void {
+    this.addSql('create type "email_address_status_enum" as enum (\'ENABLED\', \'DISABLED\', \'REQUESTED\');');
+    this.addSql('alter table "email_address" drop column "enabled";');
+
+    this.addSql('alter table "email_address" add column "status" "email_address_status_enum" not null;');
+  }
```
marode-cap, line 10: Shouldn't this also set the status on all existing rows in the table? Otherwise the `not null` constraint would be violated

src/modules/ox/domain/ox-event-handler.ts
```diff
@@ -57,6 +67,26 @@ export class OxEventHandler extends PersonenkontextCreatedEventHandler {
...
+                'bmeier',
+                '1',
+                'bmeier',
+                '1',
+                'context1',
+                'anotheremail@schule-sh.de',
+            ),
+        );
+    }*/
+
```
marode-cap, line 89: Don't forget to remove this

src/modules/keycloak-administration/domain/keycloak-user.service.ts
```diff
@@ -163,6 +163,56 @@ export class KeycloakUserService {
         }
     }
 
+    public async updateUser(
+        username: string,
+        oxUserName: OXUserName,
+        oxContextName: OXContextName,
+    ): Promise<Result<void, DomainError>> {
```
marode-cap, line 170: It this is just used to update OX-related details, it should probably have a more specific name

src/modules/keycloak-administration/domain/keycloak-user.service.ts
```diff
@@ -163,6 +163,56 @@ export class KeycloakUserService {
...
+        const foundUser: User<true> = findResult.value;
+        const newExternalSystemIDs: ExternalSystemIDs = {
+            ID_ITSLEARNING: foundUser.externalSystemIDs.ID_ITSLEARNING,
+            ID_OX: oxUserName + '@' + oxContextName,
+        };
+
+        const userRepresentation: UserRepresentation = {
+            //only attributes shall be updated here for this event
+            attributes: newExternalSystemIDs,
+        };
```
marode-cap, line 202: Be very careful here. Attributes might contain additional data, which you will just throw away here. Team 2 also uses the attributes to store the reason if a user is deactivated.

src/modules/keycloak-administration/domain/keycloak-user.service.ts
```diff
@@ -266,12 +316,18 @@ export class KeycloakUserService {
             return { ok: false, error: new KeycloakClientError('Response is invalid') };
         }
 
+        const externalSystemIDs: ExternalSystemIDs = {};
+        if (userReprDto.attributes) {
+            externalSystemIDs.ID_ITSLEARNING = userReprDto.attributes['ID_ITSLEARNING'] as string;
+            externalSystemIDs.ID_OX = userReprDto.attributes['ID_OX'] as string;
+        }
```
marode-cap, line 323: I think the attributes might actually be `string[]` (Because you can set the same attribute multiple times). Could you check if that's the case?

src/modules/keycloak-administration/domain/keycloak-client/user-representation.dto.ts
```diff
@@ -15,4 +15,7 @@ export class UserRepresentationDto {
 
     @IsNumber()
     public createdTimestamp!: number;
+
+    @IsOptional()
+    public attributes?: Record<string, string>;
```
marode-cap, line 20: As I said in the other comment, might actually be `Record<string, string[]>`

marode-cap (changes requested):

src/modules/keycloak-administration/domain/keycloak-user.service.ts
DPDS93CT: Adjustment should fix this, the ITS_LEARNING key stays as is when attribute for OX is updated.

src/modules/keycloak-administration/domain/keycloak-user.service.spec.ts
```diff
@@ -411,9 +419,48 @@ describe('KeycloakUserService', () => {
         describe('when user does not exist', () => {
             it('should return error', async () => {
                 kcUsersMock.find.mockRejectedValueOnce(new Error());
-                kcUsersMock.update.mockResolvedValueOnce();
+                //kcUsersMock.update.mockResolvedValueOnce();
```
marode-cap, line 422: Commented line

src/modules/keycloak-administration/domain/keycloak-user.service.spec.ts
```diff
@@ -411,9 +419,48 @@ describe('KeycloakUserService', () => {
...
+                );
+
+                expect(res.ok).toBeFalsy();
+            });
+        });
+
+        describe('when find returns array with undefined first element', () => {
+            it('should return error', async () => {
+                kcUsersMock.find.mockResolvedValueOnce([]);
+                //kcUsersMock.update.mockResolvedValueOnce();
```
marode-cap, line 437: Commented line

marode-cap (approved):

PR #648, opened 2024-09-09
Spsh 925

clauyan:
# Description

The 2FA-section in the frontend should only be shown to users who actually need it -> have access to a service provider, that requires 2FA. To that end I added a field to all service providers indicating whether they require 2FA or not. This is enough to decide whether or not to display the section in a users profile, but for an admin to see it (or not) in the person details, we need another way. I added the information in the 2FA (PrivacyIdea) domain, as it seemed the most logical.

[Ticket](https://ticketsystem.dbildungscloud.de/browse/SPSH-925)

## Changes

- adds a field `requires2fa` to service providers
- adds a `ServiceProviderService` to get the service providers for a person
- adds logic to `PrivacyIdeaAdministration` to determine whether the user needs 2FA

## Approval for review


src/modules/privacy-idea-administration/privacy-idea-administration.controller.ts
```diff
@@ -149,4 +151,36 @@ export class PrivacyIdeaAdministrationController {
...
+    @ApiOkResponse({
+        description: 'The requirement was successfully returned.',
+        type: TokenRequiredResponse,
+    })
+    @ApiBadRequestResponse({ description: 'A username was not given or not found.' })
+    @ApiUnauthorizedResponse({ description: 'Not authorized to get requirement information.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permissions to get requirement information.' })
+    @ApiNotFoundResponse({ description: 'Insufficient permissions to get requirement information.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while getting requirement information.' })
+    @Public()
```
marode-cap, line 166: Should not be public, since it uses PersonPermissions (Would also be the case for `getTwoAuthState`, although you didn't touch that one in this PR)

src/modules/privacy-idea-administration/privacy-idea-administration.module.ts
```diff
@@ -1,11 +1,13 @@
...
+import { Module } from '@nestjs/common';
 import { PersonModule } from '../person/person.module.js';
+import { PersonenKontextModule } from '../personenkontext/personenkontext.module.js';
+import { PrivacyIdeaAdministrationController } from './privacy-idea-administration.controller.js';
+import { PrivacyIdeaAdministrationService } from './privacy-idea-administration.service.js';
+import { ServiceProviderApiModule } from '../service-provider/service-provider-api.module.js';
 
 @Module({
-    imports: [HttpModule, PersonModule],
+    imports: [HttpModule, PersonModule, PersonenKontextModule, ServiceProviderApiModule],
```
marode-cap, line 10: Should import `ServiceProviderModule` instead of `ServiceProviderApiModule`

src/modules/service-provider/api/provider.controller.spec.ts
```diff
@@ -46,14 +45,14 @@ describe('Provider Controller Test', () => {
...
+            .overrideProvider(ServiceProviderService)
+            .useValue(createMock<ServiceProviderService>())
             .overrideProvider(ServiceProviderRepo)
             .useValue(createMock<ServiceProviderRepo>())
             .compile();
 
-        serviceProviderRepoMock = module.get(ServiceProviderRepo);
-        rolleRepoMock = module.get(RolleRepo);
+        serviceProviderServiceMock = await module.resolve<DeepMocked<ServiceProviderService>>(ServiceProviderService);
+        serviceProviderRepoMock = await module.resolve<DeepMocked<ServiceProviderRepo>>(ServiceProviderRepo);
```
marode-cap, line 55: Any reason for the switch to `.resolve`? These aren't scoped providers, are they?
If the only reason is "they are injected in a different module, so they are not accessible" you could use `.get(xxx, {strict: false})`

src/modules/service-provider/repo/service-provider.entity.ts
```diff
@@ -31,4 +31,7 @@ export class ServiceProviderEntity extends TimestampedEntity {
 
     @Property({ nullable: true })
     public keycloakRole?: string;
+
+    @Property({ nullable: true })
+    public requires2fa?: boolean;
```
marode-cap, line 36: I don't really like optional booleans, any reason why this has to be optional? The provider either requires 2fa or it doesn't

src/modules/service-provider/service-provider-api.module.ts
```diff
@@ -17,7 +18,8 @@ import { PersonModule } from '../person/person.module.js';
         RolleModule,
         PersonModule,
     ],
-    providers: [StreamableFileFactory],
+    providers: [StreamableFileFactory, ServiceProviderService],
+    exports: [ServiceProviderService],
```
marode-cap, line 22: Why the export here? If another module needs this service, it should be exported (and provided) in the `ServiceProviderModule`

marode-cap (changes requested):

src/modules/privacy-idea-administration/privacy-idea-administration.module.ts
clauyan: Fixed in [5596065](https://github.com/dBildungsplattform/dbildungs-iam-server/pull/648/commits/55960656bfa01f7dbb6cd2368742605f7e14b81c)

src/modules/service-provider/repo/service-provider.entity.ts
clauyan: I followed the rest of the properties - a lot of them are optional. But I agree that it does not really make sense here.

Fixed in [5596065](https://github.com/dBildungsplattform/dbildungs-iam-server/pull/648/commits/55960656bfa01f7dbb6cd2368742605f7e14b81c)

src/modules/service-provider/service-provider-api.module.ts
clauyan: I did not know how to deal with a Circular Dependency and this was my workaround. I referenced [this section in the docs](https://docs.nestjs.com/fundamentals/circular-dependency) and added `forwardRef` to the `RolleModule` and the `ServiceProviderModule` to fix it. Is that the right way to do it or do we have a different way?

Fixed in [5596065](https://github.com/dBildungsplattform/dbildungs-iam-server/pull/648/commits/55960656bfa01f7dbb6cd2368742605f7e14b81c)

src/modules/service-provider/api/provider.controller.spec.ts
clauyan: See my comment about the circular dependency. I think this was left over from experimentation.

Fixed in [5596065](https://github.com/dBildungsplattform/dbildungs-iam-server/pull/648/commits/55960656bfa01f7dbb6cd2368742605f7e14b81c)

src/modules/service-provider/service-provider-api.module.ts
marode-cap: If it works that's great, we had a lot of trouble with circular dependencies in the past that couldn't be solves with `forwardRef`

marode-cap (approved):

PR #668, opened 2024-09-20
Spsh 856

AlexanderUngefug:


config/dev-realm-spsh.json
marode-cap, line 1: I'll be honest, I have no idea of the contents of this file will cause any problems. I assume you've just copied the dev/prod over from the keycloak-repo and that those should also work for local development.

src/modules/authentication/types/session.d.ts
```diff
@@ -1,6 +1,11 @@
 declare module 'express-session' {
     interface SessionData {
         redirectUrl?: string;
+        passport?: {
+            user: {
+                redirect_uri?: string;
+            };
+        };
```
marode-cap, line 8: Why was this change made? Are these two different redirect URLs? 
If there is a good reason for this, please add a comment explaining it.
If there isn't a good reason this should be removed.

marode-cap (approved):

PR #673, opened 2024-09-26
SPSH-1069

DeFi3298:
Backend-Part für https://ticketsystem.dbildungscloud.de/browse/SPSH-1069


.idea/vcs.xml
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+  </component>
+</project>
```
marode-cap, line 6: This file should probably not be checked in

migrations/Migration20240828111500.ts
marode-cap, line 1: Removing migrations *might* be a bad idea. Not sure how exactly this would affect MikroORM, but it feels like this might break already running databases

migrations/Migration20240927110614.ts
```diff
@@ -0,0 +1,21 @@
...
+    async up(): Promise<void> {
+        this.addSql(
+            'create table "user_lock" ("id" uuid not null, "created_at" timestamptz not null, "updated_at" timestamptz not null, "person_id" uuid not null, "locked_by" varchar(255) null, "locked_until" timestamptz null, constraint "user_lock_pkey" primary key ("id"));',
+        );
+
+        this.addSql(
+            'alter table "user_lock" add constraint "user_lock_person_id_foreign" foreign key ("person_id") references "person" ("id") on update cascade;',
+        );
+
+        this.addSql('alter table "service_provider" add column "requires2fa" boolean not null;');
```
marode-cap, line 13: See above. Running this migration on an already deployed DB (that has already run the other migration) might crash because the 2fa-column already exists

src/modules/keycloak-administration/entity/user-lock.entity.ts
```diff
@@ -0,0 +1,15 @@
...
+@Entity({ tableName: 'user_lock' })
+export class UserLockEntity extends TimestampedEntity {
+    @ManyToOne({ entity: () => PersonEntity })
+    public readonly person!: Rel<PersonEntity>;
+
+    @Property({ nullable: true })
+    public locked_by!: string;
+
+    @Property({ nullable: true })
+    public locked_until!: Date;
```
marode-cap, line 14: Are these nullable or aren't they? The decorators and the type-definitions tell a different story

src/modules/keycloak-administration/domain/keycloak-user.service.ts
```diff
@@ -503,55 +507,52 @@ export class KeycloakUserService {
...
+    //     if (user) {
+    //         user.attributes = user.attributes ?? {};
+    //         const filteredAttributes: Record<string, string[]> = Object.fromEntries(
+    //             Object.entries(user.attributes).filter(([key]: string[]) => !(key! in LockKeys)),
+    //         );
+
+    //         user.attributes = filteredAttributes;
+    //         await kcAdminClient.users.update({ id: userId }, user);
+    //     }
+    // }
```
marode-cap, line 555: If these are no longer needed, remove the methods altogether

src/modules/keycloak-administration/domain/keycloak-user.service.ts
```diff
@@ -491,9 +494,10 @@ export class KeycloakUserService {
     }
 
     public async updateKeycloakUserStatus(
-        userId: string,
+        personId: string,
+        keyCloakUserId: string,
         enabled: boolean,
-        customAttributes?: Record<string, string>,
+        userLock: UserLock<true>,
```
marode-cap, line 500: Feels weird to pass in both an `enabled` and the `userLock`. Shouldn't the existence of the lock tell you if the user should be locked/unlocked?

src/modules/keycloak-administration/domain/user.lock.ts
```diff
@@ -0,0 +1,15 @@
+export class UserLock<WasPersisted extends boolean> {
+    private constructor(
+        public person: Persisted<string, WasPersisted>,
```
marode-cap, line 3: Use `PersonID` instead of string to signify that this is a person ID. (Also change it in `construct`)

Also, should this be using `Persisted<>`? The lock should always have a person ID, no matter if it has been saved or not. I don't see any reason for this class to use the generic at all right now

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+import { UserLockEntity } from '../entity/user-lock.entity.js';
+import { DomainError } from '../../../shared/error/domain.error.js';
+
+export function mapEntityToAggregate(entity: UserLockEntity): UserLock<true> {
+    return UserLock.construct(entity.person.id, entity.locked_by, entity.locked_until);
+}
+
+export function mapAggregateToData(userLock: UserLock<boolean>): RequiredEntityData<UserLockEntity> {
+    return {
+        person: userLock.person!,
```
marode-cap, line 16: No null-assertion. Should be unnecessary after the changes to the aggregate

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+        locked_by: userLock.locked_by,
+        locked_until: userLock.locked_until!,
```
marode-cap, line 18: No null-assertion. Should be unnecessary after the changes to the entity

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+@Injectable()
+export class UserLockRepository {
+    public readonly ROOT_ORGANISATION_ID: string;
+
+    public constructor(
+        private readonly em: EntityManager,
+        config: ConfigService<ServerConfig>,
+    ) {
+        this.ROOT_ORGANISATION_ID = config.getOrThrow<DataConfig>('DATA').ROOT_ORGANISATION_ID;
+    }
```
marode-cap, line 39: We don't need the `ROOT_ORGANISATION_ID` here. No reason to copy this from another repo

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    public async findById(id: string): Promise<Option<UserLock<true>>> {
+        const user: Option<UserLockEntity> = await this.em.findOne(UserLockEntity, { person: id });
+        if (user) {
+            return mapEntityToAggregate(user);
+        }
+        return null;
+    }
```
marode-cap, line 47: Should probably be called `findByPersonID` instead. Also use `PersonID` instead of `string`

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+        });
+        userLockEntity.assign(mapAggregateToData(userLock));
+        await this.em.persistAndFlush(userLockEntity);
+        return mapEntityToAggregate(userLockEntity);
+    }
+
+    public async deleteUserLock(personId: string): Promise<Result<void, DomainError>> {
+        await this.em.nativeDelete(UserLockEntity, { person: personId });
+        return { ok: true, value: undefined };
+    }
```
marode-cap, line 68: This can never produce an `ok: false`-result. If you want to transform any DB-errors into DomainErrors you'd need to catch the error thrown by the `nativeDelete`

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    public async update(userLock: UserLock<true>): Promise<UserLock<true> | DomainError> {
+        const userLockEntity: Loaded<UserLockEntity> = await this.em.findOneOrFail(UserLockEntity, {
+            person: userLock.person,
+        });
+        userLockEntity.assign(mapAggregateToData(userLock));
+        await this.em.persistAndFlush(userLockEntity);
+        return mapEntityToAggregate(userLockEntity);
+    }
```
marode-cap, line 63: Same as the delete-method, these would never return a DomainError (Although they might throw an error because of `persistAndFlush`).
In most other repos, we usually assume save/update methods will never fail (Or if they do, it's an unrecoverable error we can do nothing about)

src/modules/person/api/lock-user.body.params.ts
```diff
@@ -16,4 +16,11 @@ export class LockUserBodyParams {
         nullable: false,
     })
     public readonly locked_from!: string;
+
+    @IsDate()
+    @ApiProperty({
+        required: true,
+        nullable: true,
+    })
+    public readonly locked_until!: Date;
```
marode-cap, line 25: `required` and `nullable` are contradictory. 
If it really is an optional property, you should remove the `required`, add another `IsOptional`-decorator and make sure the type is `Date | undefined`

src/modules/person/api/person.controller.spec.ts
```diff
@@ -712,6 +714,7 @@ describe('PersonController', () => {
             const lockUserBodyParams: LockUserBodyParams = {
                 lock: false,
                 locked_from: '2024-01-01T00:00:00Z',
+                locked_until: new Date(),
             };
```
marode-cap, line 718: Just noticing this now: What is `locked_from` supposed to be? I assumed this was a free-form locking-reason, but here you pass it an ISO-date

src/modules/person/api/person.controller.ts
```diff
@@ -425,15 +426,17 @@ export class PersonController {
             throw new PersonDomainError(`Person with id ${personId} has no keycloak id`, personId);
         }
 
-        const lockInfo: LockInfo = {
-            lock_locked_from: lockUserBodyParams.locked_from,
-            lock_timestamp: new Date().toISOString(),
+        const userLock: UserLock<true> = {
+            person: personId,
+            locked_by: lockUserBodyParams.locked_from,
+            locked_until: new Date(lockUserBodyParams.locked_until),
```
marode-cap, line 432: `locked_until` already is a `Date`

src/modules/person/api/personenuebersicht/dbiam-personuebersicht.controller.integration.spec.ts
```diff
@@ -91,6 +92,14 @@ describe('Personenuebersicht API', () => {
...
                     useValue: createMock<PersonPermissionsRepo>(),
                 },
+                {
+                    provide: UserLockRepository,
+                    useValue: createMock<UserLockRepository>(),
+                },
+                {
+                    provide: UserLockRepository,
+                    useValue: createMock<UserLockRepository>(),
+                },
```
marode-cap, line 102: No need to do it twice

src/modules/keycloak-administration/repository/user-lock.repository.integration-spec.ts
```diff
@@ -0,0 +1,149 @@
...
+            providers: [
+                UserLockRepository,
+                {
+                    provide: ConfigService,
+                    useValue: createMock<ConfigService>({
+                        getOrThrow: jest.fn().mockReturnValue({
+                            ROOT_ORGANISATION_ID: faker.string.uuid(),
+                        }),
+                    }),
+                },
```
marode-cap, line 42: Not needed

marode-cap (changes requested):

src/modules/keycloak-administration/entity/user-lock.entity.ts
DeFi3298: adjusted

src/modules/keycloak-administration/domain/keycloak-user.service.ts
DeFi3298: removed

src/modules/person/api/lock-user.body.params.ts
DeFi3298: adjusted

.idea/vcs.xml
DeFi3298: deleted file

migrations/Migration20240828111500.ts
DeFi3298: added file back and updated snapshot

src/modules/keycloak-administration/domain/keycloak-user.service.ts
DeFi3298: Not exactly. The UserLock-Object is used here if a person is not locked to create a new lock in the database. If UserLock would not exist, it could not be created.

src/modules/keycloak-administration/domain/user.lock.ts
DeFi3298: done

src/modules/keycloak-administration/repository/user-lock.repository.ts
DeFi3298: done

src/modules/keycloak-administration/repository/user-lock.repository.ts
DeFi3298: done

src/modules/keycloak-administration/repository/user-lock.repository.ts
DeFi3298: Thanks for explaining. :)

src/modules/keycloak-administration/repository/user-lock.repository.ts
DeFi3298: removed

src/modules/person/api/personenuebersicht/dbiam-personuebersicht.controller.integration.spec.ts
DeFi3298: removed duplication

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -1,25 +1,23 @@
...
+    return UserLock.construct(entity.person.id, entity.locked_by, entity.locked_until, entity.createdAt);
 }
 
-export function mapAggregateToData(userLock: UserLock<boolean>): RequiredEntityData<UserLockEntity> {
+export function mapAggregateToData(userLock: UserLock): RequiredEntityData<UserLockEntity> {
     return {
-        person: userLock.person!,
+        person: userLock.person,
         locked_by: userLock.locked_by,
         locked_until: userLock.locked_until!,
```
marode-cap, line 16: Still has a null-assertion on `locked_until`

src/modules/person/api/lock-user.body.params.ts
```diff
@@ -15,12 +15,10 @@ export class LockUserBodyParams {
...
+    public readonly locked_by!: string;
 
+    @IsOptional()
     @IsDate()
-    @ApiProperty({
-        required: true,
-        nullable: true,
-    })
+    @ApiPropertyOptional({ description: 'Required if Befristung is set' })
     public readonly locked_until!: Date;
```
marode-cap, line 23: Optional
```suggestion
    public readonly locked_until?: Date;
```

marode-cap (approved):
Approved, but two small comments

src/modules/keycloak-administration/repository/user-lock.repository.ts
DeFi3298: done

src/modules/person/api/lock-user.body.params.ts
DeFi3298: done

marode-cap (approved):

PR #682, opened 2024-10-01
SPSH-1121 Add manual sync to external systems (specifically itslearning)

marode-cap:
# Description
Create mechanism to synchronize persons with external systems

## Approval for review


src/modules/person/api/person.controller.ts
```diff
@@ -441,6 +444,30 @@ export class PersonController {
...
+        @Permissions() permissions: PersonPermissions,
+    ): Promise<void> {
+        const personResult: Result<Person<true>> = await this.personRepository.getPersonIfAllowed(
+            personId,
+            permissions,
+            // TODO SPSH-1136 Update the RollenSystemRecht
+            [RollenSystemRecht.PERSONEN_VERWALTEN],
+        );
+        if (!personResult.ok) {
+            throw new NotFoundOrNoPermissionError(personId);
```
YoussefBouch, line 465: Is it not better to distinguish errors here? Maybe return a not found if findById for Person doesn't find anything and NoPermission if the admin has no permission to find the user?

src/modules/person/api/person.controller.ts
```diff
@@ -441,6 +444,30 @@ export class PersonController {
...
+        }
+
+        this.eventService.publish(new PersonExternalSystemsSyncEvent(personId));
```
YoussefBouch, line 468: As far as I know, events should be published in the domain/persistence instead of api layer? Is it good practice for the eventService to be injected in a controller?

src/modules/itslearning/repo/role-utils.ts
```diff
@@ -0,0 +1,52 @@
...
+    );
+}
+
+export function rollenartToIMSESInstitutionRole(role: RollenArt): IMSESInstitutionRoleType {
+    return ROLLENART_TO_ITSLEARNING_ROLE[role];
+}
+
+export function rollenartToIMSESRole(role: RollenArt): IMSESRoleType {
+    return ROLLENART_TO_IMSES_ROLE[role];
+}
```
YoussefBouch, line 52: Not important but since this is a utils file, I think it's nice to see what every method is supposed to do (Even though those are not very hard to read).

YoussefBouch (commented):
Shouldn't we handle the case here where the person is deleted by another admin and the second admin tries to sync the same person... 

Currently we get a 404 with an ugly error message in the frontend. I'd prefer if we sent a 18nKey like we always do and let the frontend display it in a good way. What do you think?

src/modules/person/api/person.controller.ts
marode-cap: If we differentiate, a user could figure out if a person with a specific ID exists. Of course just knowing that an ID exists is not really "sensitive" information, but we do this in multiple places to not leak the information

marode-cap:
> Shouldn't we handle the case here where the person is deleted by another admin and the second admin tries to sync the same person...
> 
> Currently we get a 404 with an ugly error message in the frontend. I'd prefer if we sent a 18nKey like we always do and let the frontend display it in a good way. What do you think?

I don't think we can know (or should care) as the backend, if the person was deleted in the past. If the person is not in the database, we return a 404.

Also, we do return an i18nKey, so this might be a problem in the frontend?

src/modules/person/api/person.controller.ts
marode-cap: Going to talk with an architect

marode-cap: Should not create an extra layer just for the event

YoussefBouch (approved):

PR #685, opened 2024-10-07
SPSH-1146

YoussefBouch:
# Description

We had the problem that the limit was preventing for example the Schuladmin from seeing his assigned Schule if the Schule was outside the 25 limit. 

Solution: We retrieve the permitted orgas first and then we query all orgas using those Ids. If the admin has permission on all orgas then we don't query used permitted orgas since its optional.

## Approval for review


src/modules/personenkontext/domain/personenkontext-workflow.spec.ts
```diff
@@ -465,6 +465,69 @@ describe('PersonenkontextWorkflow', () => {
...
+                orgaIds: permittedOrgaIds, // Only permit Org 1
+            });
+
+            const result: Organisation<true>[] = await anlage.findAllSchulstrukturknoten(
+                personpermissionsMock,
+                undefined,
+            );
+
+            // Ensure only the permitted organisation is returned
+            expect(result.length).toEqual(1); // Only Org 1 should be returned
```
marode-cap, line 529: Maybe also check that the returned orga is actually the orga we expect, not just that the length is 1

marode-cap (approved):

PR #689, opened 2024-10-08
SPSH-814

DeFi3298:
https://ticketsystem.dbildungscloud.de/browse/SPSH-814


src/modules/cron/cron.module.ts
```diff
@@ -0,0 +1,11 @@
+import { HttpModule } from '@nestjs/axios';
+import { Module } from '@nestjs/common';
+import { CronController } from './cron.controller.js';
+import { KeycloakAdministrationModule } from '../keycloak-administration/keycloak-administration.module.js';
+import { PersonModule } from '../person/person.module.js';
+
+@Module({
+    imports: [HttpModule, PersonModule, KeycloakAdministrationModule],
```
marode-cap, line 8: I don't think the `HttpModule` is needed

src/modules/person/persistence/person.repository.ts
```diff
@@ -629,4 +629,26 @@ export class PersonRepository {
...
+            $and: [
+                { personalnummer: { $eq: null } },
+                {
+                    personenKontexte: {
+                        createdAt: { $lte: daysAgo }, //Check that createdAt is older than 56 days
+                        rolleId: {
+                            merkmale: { merkmal: RollenMerkmal.KOPERS_PFLICHT },
+                        },
+                    },
+                },
```
marode-cap, line 647: This *might* need to have a `$some` condition, not sure what the default is when filtering across collections.
Could you make sure this works (Maybe even create a integration test for this case), by creating a person with two personenkontexte:
- One that would qualify the person for locking (Has the kopers-pflicht and is expired)
- One that would not lock the person (Has kopers-pflicht but is not expired)

Anyway, `$some` would make the intention more clear

```typescript
{
    personenKontexte: {
        $some: {
            createdAt: { $lte: daysAgo }, //Check that createdAt is older than 56 days
            rolleId: {
                merkmale: { merkmal: RollenMerkmal.KOPERS_PFLICHT },
            },
        },
    },
},
```

src/modules/cron/cron.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+    @ApiUnauthorizedResponse({ description: 'Not authorized to lock user.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permissions to lock user.' })
+    @ApiNotFoundResponse({ description: 'Insufficient permissions to lock user.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while trying to lock user.' })
+    public async KoPersUserLock(): Promise<boolean> {
+        try {
+            const keyCloakIds: string[] = await this.personRepository.getKoPersUserLockList();
+            if (keyCloakIds.length === 0) {
+                return false;
+            }
```
marode-cap, line 40: Should this be `false`? In my opinion, locking no users because there aren't any to lock, is still a success.

src/modules/cron/cron.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+            const allSuccessful: boolean = results.every(
+                (result: PromiseSettledResult<Result<void, DomainError>>) =>
+                    result.status === 'fulfilled' && result.value.ok === true,
+            );
+
+            if (allSuccessful) {
+                return true;
+            } else {
+                return false;
+            }
```
marode-cap, line 55: Just do
```typescript
return allSuccessful;
```

src/modules/cron/cron.controller.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+                {
+                    provide: KeycloakUserService,
+                    useValue: createMock<KeycloakUserService>(),
+                },
+                {
+                    provide: PersonRepository,
+                    useValue: createMock<PersonRepository>(),
+                },
+                CronController,
+            ],
```
marode-cap, line 25: `CronController` should not be in `providers`, put it into `controllers`

src/modules/cron/cron.controller.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    beforeEach(() => {
+        jest.resetAllMocks();
+    });
+
+    describe('/PUT cron/kopers-lock', () => {
+        describe('when there are users to lock', () => {
+            it('should return true when all users are successfully locked', async () => {
+                const mockKeycloakIds: string[] = ['user1', 'user2', 'user3'];
+                personRepositoryMock.getKoPersUserLockList.mockResolvedValueOnce(mockKeycloakIds);
+                keycloakUserServiceMock.updateKeycloakUserStatus.mockResolvedValue({ ok: true, value: undefined });
```
marode-cap, line 42: Make this explicit
```suggestion
                keycloakUserServiceMock.updateKeycloakUserStatus.mockResolvedValueOnce({ ok: true, value: undefined });
                keycloakUserServiceMock.updateKeycloakUserStatus.mockResolvedValueOnce({ ok: true, value: undefined });
                keycloakUserServiceMock.updateKeycloakUserStatus.mockResolvedValueOnce({ ok: true, value: undefined });
```

marode-cap (changes requested):

marode-cap (approved):

PR #691, opened 2024-10-09
Spsh 909

tikle-cap:
https://ticketsystem.dbildungscloud.de/browse/SPSH-909

## Approval for review


src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
```diff
@@ -90,15 +89,19 @@ export class PrivacyIdeaAdministrationService {
...
             this.jwtToken = response.data.result.value.token;
             return this.jwtToken;
         } catch (error) {
-            if (error instanceof Error) {
+            if (error instanceof PIUnavailableError) {
+                throw error;
+            } else if (error instanceof Error) {
                 throw new Error(`Error fetching JWT token: ${error.message}`);
             } else {
                 throw new Error(`Error fetching JWT token: Unknown error occurred`);
```
marode-cap, line 107: I would have moved all the error mapping into the `catchError` pipe, that way it would be cleaner and the entire try-catch would be unnecessary (Not just here, but all the try-catches in this file).
But this works, so I'll leave that up to you

marode-cap (approved):

PR #693, opened 2024-10-10
Spsh 1066

AlexanderUngefug:


src/modules/privacy-idea-administration/privacy-idea-administration.service.ts
```diff
@@ -436,4 +440,33 @@ export class PrivacyIdeaAdministrationService {
...
+            }
+        }
+    }
+
+    @EventHandler(PersonDeletedEvent)
+    public async handlePersonDeletedEvent(event: PersonDeletedEvent): Promise<void> {
+        this.logger.info(`Received PersonDeletedEvent, personId:${event.personId}`);
+        const userTokens: PrivacyIdeaToken[] = await this.getUserTokens(event.referrer);
+        if (userTokens !== undefined && userTokens.length > 0) {
+            await this.resetToken(event.referrer);
```
marode-cap, line 468: `this.getUserTokens` should never return undefined

src/modules/privacy-idea-administration/privacy-idea-administration.service.spec.ts
```diff
@@ -832,4 +837,93 @@ describe(`PrivacyIdeaAdministrationService`, () => {
             expect(result).toBe(requires2fa);
         });
     });
+
+    describe('deleteUser', () => {
```
marode-cap, line 841: names of the describe-blocks are switched around (the first one is testing the event handler)

src/modules/privacy-idea-administration/privacy-idea-administration.service.spec.ts
```diff
@@ -832,4 +837,93 @@ describe(`PrivacyIdeaAdministrationService`, () => {
...
+
+        beforeEach(() => {
+            personId = faker.string.uuid();
+            referrer = faker.string.alpha();
+            emailAddress = faker.internet.email();
+            event = new PersonDeletedEvent(personId, referrer, emailAddress);
+            jest.spyOn(
+                service as unknown as { resetToken: (username: string) => Promise<ResetTokenResponse> },
+                'resetToken',
+            ).mockResolvedValueOnce(mockResetTokenResponse);
```
marode-cap, line 894: There are a lot of `spyOn`s here, that aren't actually used.
`deleteUser` should not need any spy, since it doesn't call function of the service.

And for `handlePersonDeletedEvent`: I get why you use the spyOn, but I'd recommend moving `handlePersonDeletedEvent` into its own file (I usually separate event handlers from other logic). That way you could use our usual mocking of dependencies instead of spying on the system under test

src/modules/privacy-idea-administration/privacy-idea-administration.service.spec.ts
```diff
@@ -301,13 +304,15 @@ describe(`PrivacyIdeaAdministrationService`, () => {
                 { provide: HttpService, useValue: createMock<HttpService>() },
                 { provide: ServiceProviderService, useValue: createMock<ServiceProviderService>() },
                 { provide: PersonenkontextService, useValue: createMock<PersonenkontextService>() },
+                { provide: ClassLogger, useValue: createMock<ClassLogger>() },
```
marode-cap, line 307: Import the `LoggingTestModule` instead

marode-cap (changes requested):

marode-cap (approved):

PR #695, opened 2024-10-10
SPSH-1020: update realm configuration for NextCloud client credentials

niklaskoopmann:
# Description

Configuration to connect the Online-Dateiablage (NextCloud) as Client and Service Provider.

## Links to Tickets or other PRs

- https://ticketsystem.dbildungscloud.de/browse/SPSH-1020

## Approval for review


config/dev-realm-spsh.json
```diff
@@ -103,6 +103,7 @@
       }
     ],
     "client": {
+      "nextcloud-bildungsablage-test": [],
```
marode-cap, line 106: Should this be here?

config/dev-realm-spsh.json
```diff
@@ -1237,6 +1238,60 @@
...
+        "adminUrl": "",
+        "baseUrl": "",
+        "surrogateAuthRequired": false,
+        "enabled": true,
+        "alwaysDisplayInConsole": false,
+        "clientAuthenticatorType": "client-secret",
+        "secret": "${KC_NEXTCLOUD_CLIENT_SECRET}",
+        "redirectUris": [
+          "https://sh.lumaserv.dev/*"
+        ],
```
marode-cap, line 1257: Usually you'd set the `rootUrl` to the URL (i.e. `https://sh.lumaserv.dev`) and set `redirectUris` to relative paths (in this case `/*`. Speaking of which, can you narrow this down further instead of using the wildcard?)

config/dev-realm-spsh.json
```diff
@@ -1237,6 +1238,60 @@
...
+        "webOrigins": [
+          "/*"
+        ],
+        "notBefore": 0,
+        "bearerOnly": false,
+        "consentRequired": false,
+        "standardFlowEnabled": true,
+        "implicitFlowEnabled": true,
+        "directAccessGrantsEnabled": true,
```
marode-cap, line 1266: Disable the implicit flow (unless its needed) and definitely disable the direct access grants

config/dev-realm-spsh.json
marode-cap: Was asking, because I believe this would be the client-id.
But you use ${KC_NEXTCLOUD_CLIENT_ID} to set the ID for the client, both of them should probably match

niklaskoopmann: Oh, yeah, you're right, thanks. Completely overlooked that... I'll fix that and look into your other comments on Monday. 👍

PR #728, opened 2024-10-29
Spsh 941

AlexanderUngefug:


src/modules/authentication/passport/oidc.strategy.ts
```diff
@@ -26,11 +26,36 @@ export function extractStepUpLevelFromJWT(jwt: string | undefined): StepUpLevel
...
 }
 
+export function getLowestStepUpLevel(): StepUpLevel {
+    return StepUpLevel.SILVER;
+}
+
+// timeout in seconds
+export function isStepUpTimeOver(req: Request, timeout: number): boolean {
+    const currentTime: number = new Date().getTime();
+    if (!req.session?.lastRouteChangeTime) return false;
```
marode-cap, line 40: Could use `Date.now()`, but doesn't really matter (Performance is a tiny bit better)

src/modules/person/domain/required-step-up-level-not-met.error.ts
```diff
@@ -0,0 +1,7 @@
+import { PersonDomainError } from './person-domain.error.js';
+
+export class RequiredStepUpLevelNotMetError extends PersonDomainError {
+    public constructor(entityId: string | undefined, details?: unknown[] | Record<string, undefined>) {
+        super(`The action could not be perfomed, bequase the required step up level was not met`, entityId, details);
```
marode-cap, line 5: I don't think the parameter for `entityId` is used anywhere. Do we even need it? Also, should this be a PersonDomainError? Feels a lot more like an `AuthenticationDomainError`

src/shared/config/config.env.ts
```diff
@@ -5,9 +5,10 @@ import { HostConfig } from './host.config.js';
 import { ItsLearningConfig } from './itslearning.config.js';
 import { LdapConfig } from './ldap.config.js';
 import { PrivacyIdeaConfig } from './privacyidea.config.js';
+import { SystemConfig } from './system.config.js';
 import { OxConfig } from './ox.config.js';
 
-export default (): {
+export type Config = {
     DB: Partial<DbConfig>;
```
marode-cap, line 12: Good idea to split this off, but I don't think we need to export this type

src/shared/config/config.loader.spec.ts
```diff
@@ -105,13 +109,18 @@ describe('configloader', () => {
...
                 const validatedConfig: JsonConfig = loadConfigFiles();
                 expect(validatedConfig).toBeInstanceOf(JsonConfig);
                 expect(readFileSyncSpy).toHaveBeenCalledTimes(2);
             });
         });
 
         describe('When there is no secrets config', () => {
+            beforeAll(() => {
+                jest.clearAllMocks();
+            });
```
marode-cap, line 122: Was this change needed? I don't think any changes you did should affect these tests

src/modules/person/api/person.controller.ts
```diff
@@ -101,6 +103,7 @@ export class PersonController {
     }
 
     @Post()
+    @UseGuards(StepUpGuard)
     @HttpCode(HttpStatus.CREATED)
```
marode-cap, line 107: General comment for all `UseGuards(StepUpGuard)`:
You can decorate the entire controller-class, to add the guard to all routes (only if all of the routes in that controller need it, of course)

src/modules/authentication/passport/oidc.strategy.spec.ts
```diff
@@ -145,13 +152,99 @@ describe('OpenIdConnectStrategy', () => {
...
-            request.query['requiredStepUpLevel'] = StepUpLevel.GOLD;
+            request.session.requiredStepupLevel = StepUpLevel.GOLD;
             sut.authenticate(request);
 
             expect(superPassportSpy).toHaveBeenCalledWith(request, { acr_values: 'gold' });
         });
     });
 
+    const mockTime = (time: number): jest.SpyInstance => jest.spyOn(Date.prototype, 'getTime').mockReturnValue(time);
+
```
marode-cap, line 163: If possible, I'd prefer using jests fake-timers instead of mocking the Date object

marode-cap (changes requested):

src/shared/config/config.env.ts
AlexanderUngefug: i need the type for the unit tests

charts/dbildungs-iam-server/config/config.json
marode-cap, line 1: Please reformat to match existing indentation. It's hard to miss relevant changes if everything is marked as a change

src/modules/person/domain/required-step-up-level-not-met.error.ts
```diff
@@ -1,7 +1,7 @@
-import { PersonDomainError } from './person-domain.error.js';
+import { AuthenticationDomainError } from '../../authentication/domain/authentication-domain.error.js';
 
-export class RequiredStepUpLevelNotMetError extends PersonDomainError {
-    public constructor(entityId: string | undefined, details?: unknown[] | Record<string, undefined>) {
-        super(`The action could not be perfomed, bequase the required step up level was not met`, entityId, details);
+export class RequiredStepUpLevelNotMetError extends AuthenticationDomainError {
+    public constructor() {
+        super(`The action could not be perfomed, bequase the required step up level was not met`, undefined, undefined);
     }
```
marode-cap, line 6: The optional "details" parameter would have been fine in my opinion, we sometimes use it to pass a stacktrace to the error. But probably won't matter in this case, since it's just an API error

marode-cap (approved):

src/modules/person/domain/required-step-up-level-not-met.error.ts
AlexanderUngefug: details are added back

PR #734, opened 2024-11-03
SPSH-1264

YoussefBouch:
# Description
**Deleting a user now only delete it's record in the userLock table. This should prevent the error while deleting a blocked user.**

## Approval for review


src/modules/keycloak-administration/entity/user-lock.entity.ts
```diff
@@ -1,10 +1,17 @@
...
 import { PersonEntity } from '../../person/persistence/person.entity.js';
 
 @Entity({ tableName: 'user_lock' })
 export class UserLockEntity extends TimestampedEntity {
-    @ManyToOne({ entity: () => PersonEntity })
+    @ManyToOne({
+        entity: () => PersonEntity,
+        fieldName: 'person_id',
+        columnType: 'uuid',
+        cascade: [Cascade.REMOVE],
```
marode-cap, line 11: I belive this should not be set to `Cascade.REMOVE`.
If a UserLockEntity gets removed, it should not remove the user.

marode-cap (changes requested):

src/modules/keycloak-administration/entity/user-lock.entity.ts
YoussefBouch: Me and @pkleybolte talked about this when implementing cascading for Personenkontext with Person and cascading only works this way smh. Deleting userLocks doesn't delete Person here anyways (Tried it many times) and removing this part of the code in the user-lock.entity.ts persists the records there even when the person is deleted.

marode-cap: Weird. Reading the MikroORM docs I came to the other conclusion, but if you've tested and confirmed it thats fine by me.

marode-cap (approved):

PR #746, opened 2024-11-11
SPSH-1143

DeFi3298:
Ticket: https://ticketsystem.dbildungscloud.de/browse/SPSH-1143


src/modules/person/api/person.controller.ts
```diff
@@ -526,6 +528,21 @@ export class PersonController {
...
+            if (koperslock) {
+                await this.keycloakUserService.updateKeycloakUserStatus(
+                    params.personId,
+                    result.keycloakUserId!,
+                    koperslock,
+                    false,
+                );
+            }
+        }
+
```
marode-cap, line 545: Could we move this into the `PersonRepository.updatePersonMetadata`? If we update the kopers somewhere else the users would not be unlocked (I don't know why we would do that, but could happen in the future. Maybe bulk-operations or something like that)

marode-cap (changes requested):

src/modules/person/api/person.controller.ts
DeFi3298: all done

marode-cap (approved):

PR #751, opened 2024-11-13
SPSH-1113

DeFi3298:
Ticket: https://ticketsystem.dbildungscloud.de/browse/SPSH-1113

### DIFF
--- a/src/modules/cron/cron.controller.ts
+++ b/src/modules/cron/cron.controller.ts
@@ -193,7 +193,7 @@ export class CronController {
     @ApiForbiddenResponse({ description: 'Insufficient permissions to unlock users.' })
     @ApiNotFoundResponse({ description: 'Insufficient permissions to unlock users.' })
     @ApiInternalServerErrorResponse({ description: 'Internal server error while trying to unlock users.' })
-    public async unlockUsersAfterTimeout(@Permissions() permissions: PersonPermissions): Promise<boolean> {
+    public async unlockUsersWithExpiredLocks(@Permissions() permissions: PersonPermissions): Promise<boolean> {
         try {
             const userLocks: UserLock[] = await this.userLockRepository.getLocksToUnlock();
             if (userLocks.length === 0) {
@@ -212,8 +212,8 @@ export class CronController {
                     return this.keyCloakUserService.updateKeycloakUserStatus(
                         person.value.id,
                         person.value.keycloakUserId,
-                        true,
                         userLock,
+                        false,
                     );
                 }),
             );
--- a/src/modules/cron/cron.controller.ts
+++ b/src/modules/cron/cron.controller.ts
@@ -5,6 +5,7 @@ import {
     ApiUnauthorizedResponse,
     ApiForbiddenResponse,
     ApiNotFoundResponse,
+    ApiOkResponse,
     ApiInternalServerErrorResponse,
     ApiBearerAuth,
     ApiOAuth2,
@@ -23,6 +24,9 @@ import { PersonenkontextWorkflowFactory } from '../personenkontext/domain/person
 import { Personenkontext } from '../personenkontext/domain/personenkontext.js';
 import { PersonenkontexteUpdateError } from '../personenkontext/domain/error/personenkontexte-update.error.js';
 import { PersonID } from '../../shared/types/aggregate-ids.types.js';
+import { UserLockRepository } from '../keycloak-administration/repository/user-lock.repository.js';
+import { Person } from '../person/domain/person.js';
+import { EntityNotFoundError } from '../../shared/error/entity-not-found.error.js';
 
 @Controller({ path: 'cron' })
 @ApiBearerAuth()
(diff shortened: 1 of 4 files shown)


src/modules/cron/cron.controller.ts
```diff
@@ -178,4 +183,49 @@ export class CronController {
...
+            }
+
+            const results: PromiseSettledResult<Result<void, DomainError>>[] = await Promise.allSettled(
+                userLocks.map(async (userLock: UserLock) => {
+                    const person: Result<Person<true>> = await this.personRepository.getPersonIfAllowed(
+                        userLock.person,
+                        permissions,
+                    );
+                    if (!person.ok || !person.value.keycloakUserId) {
+                        return { ok: false, error: new EntityNotFoundError() };
```
YoussefBouch, line 210: I was gonna request a change here to return a more descriptive error and provide feedback on which user can't be unlocked but seems like all the endpoints just throw a generic error and say it's an internal server error... Not so good but I'll just approve it since it's done this way everywhere and this is only used for cron jobs.

src/modules/keycloak-administration/repository/user-lock.repository.ts
```diff
@@ -62,4 +62,15 @@ export class UserLockRepository {
     public async deleteUserLock(personId: string, lockOccasion: PersonLockOccasion): Promise<void> {
         await this.em.nativeDelete(UserLockEntity, { person: personId, locked_occasion: lockOccasion });
     }
+
+    public async getLocksToUnlock(): Promise<UserLock[]> {
+        const today: Date = new Date();
+
+        const filters: QBFilterQuery<UserLockEntity> = {
+            locked_until: { $lte: today },
```
YoussefBouch, line 70: Are we sure about this? The saved lockedUntil in the table userLock is in UTC and not in german time. Here we compare the german time (which is UTC + 1 )  to the UTC time.  

We could maybe use ` const now = DateTime.utc();` instead.

@kristoff-kiefer  @marode-cap What do you think?

src/modules/cron/cron.controller.ts
```diff
@@ -178,4 +183,49 @@ export class CronController {
...
+    @ApiNotFoundResponse({ description: 'Insufficient permissions to unlock users.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while trying to unlock users.' })
+    public async unlockUsersWithExpiredLocks(@Permissions() permissions: PersonPermissions): Promise<boolean> {
+        try {
+            const userLocks: UserLock[] = await this.userLockRepository.getLocksToUnlock();
+            if (userLocks.length === 0) {
+                return true;
+            }
+
+            const results: PromiseSettledResult<Result<void, DomainError>>[] = await Promise.allSettled(
```
YoussefBouch, line 203: I guess we use Promise.allSettled instead of Promise.all here because we still want to unlock another user even though the previous one/ones fail but apparently the endpoint only returns true/false or throws a generic error so we don't really handle the resolved promises individually for example count the number of users that resolved or failed and log them for metrics... But I guess this is something we could add later for all the endpoints here...

src/modules/cron/cron.controller.ts
```diff
@@ -178,4 +183,49 @@ export class CronController {
...
+                    return this.keyCloakUserService.updateKeycloakUserStatus(
+                        person.value.id,
+                        person.value.keycloakUserId,
+                        userLock,
+                        false,
+                    );
+                }),
+            );
+
+            const allSuccessful: boolean = results.every(
```
YoussefBouch, line 221: I guess we use Promise.allSettled instead of Promise.all here because we still want to unlock another user even though the previous one/ones fail but apparently the endpoint only returns true if all promises resolve successfully so we might get an error even though we unlocked 1000 users and only 1 was errorful... Not sure if that's good and helpful but we might need to refactor all these endpoints if we want more informations and metrics in the future.

src/modules/cron/cron.controller.ts
DeFi3298: The reasoning here was, that this code is only used for cron jobs. It might be a good idea, that we extend it in the future, but right now it's good to just get the functionality with prio.

src/modules/keycloak-administration/repository/user-lock.repository.ts
marode-cap: Pretty sure `DateTime` is not a JS-API. When calling the `Date`-constructor without any arguments, the internal timestamp will be the same as if I used `Date.now()`, which is in UTC.
I think MikroORM only compared the actual timestamp, but I couldn't say for sure

YoussefBouch: @DeFi3298 Can you maybe check what "today" is according to MikroORM? To make sure we are actually comparing UTC with UTC.

DeFi3298: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date

"When no parameters are provided, the newly-created Date object represents the current date and time as of the time of instantiation. The returned date's timestamp is the same as the number returned by Date.now()"

YoussefBouch (approved):

PR #767, opened 2024-11-19
Dbp 1081 dbildungs iam server took too long to get in ready state

M-Schiborr:
Backend Pod is now faster green and in ready state

### DIFF
--- a/charts/dbildungs-iam-server/values.yaml
+++ b/charts/dbildungs-iam-server/values.yaml
@@ -70,8 +70,8 @@ backend:
       memory: 200Mi
   startupProbe:
     enabled: true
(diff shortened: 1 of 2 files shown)


YannickEvers (approved):

charts/dbildungs-iam-server/values.yaml
```diff
@@ -70,8 +70,8 @@ backend:
       memory: 200Mi
   startupProbe:
     enabled: true
-    initialDelaySeconds: 60
-    periodSeconds: 60
+    initialDelaySeconds: 30
```
aimee-889, line 73: could we further decrease this to 10? it seems to work and would be even faster

PR #777, opened 2024-11-21
SPSH-1275

DeFi3298:
Ticket: https://ticketsystem.dbildungscloud.de/browse/SPSH-1275


tikle-cap (approved):

he-meyer (approved):

src/modules/person/persistence/person.repository.integration-spec.ts
```diff
@@ -113,6 +115,10 @@ describe('PersonRepository Integration', () => {
                     provide: KeycloakUserService,
                     useValue: createMock<KeycloakUserService>(),
                 },
+                {
+                    provide: ClassLogger,
+                    useValue: createMock<ClassLogger>(),
+                },
```
marode-cap, line 121: Import LoggingTestModule

src/modules/personenkontext/api/dbiam-personenkontext-workflow.controller.ts
```diff
@@ -243,23 +246,47 @@ export class DbiamPersonenkontextWorkflowController {
             params.personalnummer || undefined,
             params.befristung || undefined,
         );
+        if (savedPersonWithPersonenkontext instanceof Error) {
+            this.logger.error(
```
marode-cap, line 250: `instanceof Error` can have very weird behaviour in JS. Can we switch this to `!(savedPersonWithPersonenkontext instanceof PersonPersonenkontext)`?

src/modules/personenkontext/specification/personenkontext-specifications.integration-spec.ts
```diff
@@ -99,6 +100,10 @@ describe('PersonenkontextSpecifications Integration', () => {
                     provide: RolleRepo,
                     useValue: createMock<RolleRepo>(),
                 },
+                {
+                    provide: ClassLogger,
+                    useValue: createMock<ClassLogger>(),
+                },
```
marode-cap, line 106: Import LoggingTestModule

src/modules/privacy-idea-administration/privacy-idea-administration.controller.spec.ts
```diff
@@ -52,6 +54,10 @@ describe('PrivacyIdeaAdministrationController', () => {
                     provide: PersonRepository,
                     useValue: createMock<PersonRepository>(),
                 },
+                {
+                    provide: ClassLogger,
+                    useValue: createMock<ClassLogger>(),
+                },
```
marode-cap, line 60: Import LoggingTestModule

marode-cap (changes requested):

src/modules/personenkontext/api/dbiam-personenkontext-workflow.controller.ts
DeFi3298: Wont work this way.
"The instanceof operator expects the right-hand operand to be a class or a constructor function, but here PersonPersonenkontext is likely a TypeScript interface or type alias, which is purely a compile-time construct and not available at runtime."

marode-cap: Ah, I thought it already was a class. In that case keep it like this, although I don't like how were returning multiple different things here (This should really be a `Result<>` instead, but it was already like this)

marode-cap (approved):

PR #792, opened 2024-11-28
SPSH-1519 - StepUpGuard fix for direct grant flow

AlexanderUngefug:


casparneumann-cap:
PO/ Architect Decision: Will not be used. Lines will be commented out for migration

PR #804, opened 2024-12-02
SPSH-1512: add USER_ID attribute to Keycloak users

niklaskoopmann:
Required by NextCloud, see https://ticketsystem.dbildungscloud.de/browse/SPSH-1512

### DIFF
--- a/src/console/dbseed/domain/db-seed.service.spec.ts
+++ b/src/console/dbseed/domain/db-seed.service.spec.ts
@@ -440,6 +440,7 @@ describe('DbSeedService', () => {
                     'test@example.com',
                     faker.date.recent(),
                     {
+                        USER_ID: [faker.string.uuid()],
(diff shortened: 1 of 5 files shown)


marode-cap (approved):
If this is only used for Nextcloud, maybe name the attribute differently. Otherwise fine by me

niklaskoopmann:
> If this is only used for Nextcloud, maybe name the attribute differently. Otherwise fine by me

@marode-cap Thanks for pointing that out. I had originally planned to use it as a general attribute to be used by client-specific mappers as needed. After discussing this with @kristoff-kiefer last week we decided to keep this and the ItsLearning ID attribute client-specific since this solution is only temporary and will be replaced by the Keycloak plugin we are desperately waiting to find time to implement...

Long story short, I renamed it now.

PR #818, opened 2024-12-05
added 2nd ingress

mcpovel:
Add secondary ingress if enabled


dimapin (approved):
Replacing 2nd with second would be better

kristoff-kiefer (approved):

PR #821, opened 2024-12-05
Dbp 1156 create pod disruption budget for spsh applications

M-Schiborr:
# Description

## Approval for review

### DIFF
--- a/charts/dbildungs-iam-server/templates/pdb.yaml
+++ b/charts/dbildungs-iam-server/templates/pdb.yaml
@@ -19,5 +19,4 @@ spec:
       - dbildungs-iam-server
       - schulportal-client
       - dbildungs-iam-keycloak
-      - dbildungs-iam-ldap
 {{- end }}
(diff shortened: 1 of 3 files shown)


sahassou (approved):

charts/dbildungs-iam-server/templates/pdb.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+spec:
+  minAvailable: {{ .Values.podDisruptionBudget.minAvailable }}
+  selector:
+    matchExpressions:
+    - key: app.kubernetes.io/name
+      operator: In
+      values:
+      - dbildungs-iam-server
```
simoncolincap, line 19: Wouldn't this mean that only one of any of the 3 pods should be available? I think it would be better to have a PDB for each component

M-Schiborr: true it actually sums all available pods together, i will create on pdb for each component

simoncolincap (approved):

PR #836, opened 2024-12-10
Spsh 1378

clauyan:
# Description

- adds an option for the node runtime to ensure timely garbage collection to avoid OOM-errors
- fix double indentation so `extraEnvVars` works

## Approval for review

### DIFF
--- a/charts/dbildungs-iam-server/values.yaml
+++ b/charts/dbildungs-iam-server/values.yaml
@@ -130,7 +130,9 @@ backend:
     interval: 30m
     endpoints:
       - port: 'web'
-  extraEnvVars: []
+  extraEnvVars:
+    - name: NODE_OPTIONS
+      value: "--max-old-space-size=3584"
   extraVolumes: []
   volumeMounts:
     - mountPath: /app/config/config.json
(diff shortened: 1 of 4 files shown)


charts/dbildungs-iam-server/values.yaml
```diff
@@ -124,7 +124,9 @@ backend:
     interval: 30m
     endpoints:
       - port: 'web'
-  extraEnvVars: []
+  extraEnvVars:
```
aimee-889, line 127: is there a reason why you use this extraEnvVars? I think this setting would be easier to understand if this value is also placed in the configmap. I think i myself added the extraEnvVars, even though i can currently not think of a reason why we would need it :D Might get deleted as well

clauyan: It made sense to me to have it _separate_, because the setting is fairly low-level/does not have much to do with the app itself and probably won't be changed often.
I follow you reasoning though. I don't mind either way, so I will move it into the ConfigMap :)

aimee-889 (approved):

clauyan:
Pushed back to next version; Waiting for new base branch

PR #839, opened 2024-12-11
SPSH-1626 itslearning correlation-IDs

marode-cap:
# Description

## Approval for review

### DIFF
--- a/src/modules/itslearning/event-handlers/itslearning-persons.event-handler.ts
+++ b/src/modules/itslearning/event-handlers/itslearning-persons.event-handler.ts
@@ -43,58 +43,71 @@ export class ItsLearningPersonsEventHandler {
     @EventHandler(PersonRenamedEvent)
     public async personRenamedEventHandler(event: PersonRenamedEvent): Promise<void> {
         await this.personUpdateMutex.runExclusive(async () => {
-            this.logger.info(`Received PersonRenamedEvent, ${event.personId}`);
+            this.logger.info(`[EventID: ${event.eventID}] Received PersonRenamedEvent, ${event.personId}`);
 
             if (!this.ENABLED) {
-                return this.logger.info('Not enabled, ignoring event.');
+                return this.logger.info(`[EventID: ${event.eventID}] Not enabled, ignoring event.`);
             }
 
             if (!event.referrer) {
-                return this.logger.error(`Person with ID ${event.personId} has no username!`);
+                return this.logger.error(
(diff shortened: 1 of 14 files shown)


src/modules/itslearning/event-handlers/itslearning-persons.event-handler.ts
```diff
@@ -43,69 +43,82 @@ export class ItsLearningPersonsEventHandler {
...
             if (!event.referrer) {
-                return this.logger.error(`Person with ID ${event.personId} has no username!`);
+                return this.logger.error(
+                    `[EventID: ${event.eventID}] Person with ID ${event.personId} has no username!`,
+                );
             }
 
             const readPersonResult: Option<PersonResponse> = await this.itslearningPersonRepo.readPerson(
                 event.personId,
+                `${event.eventID}-PERSON-EXISTS-CHECK`,
```
clauyan, line 60: Is it possible to put this name + formatting in the base-event, so it is centralized?

clauyan (approved):
Do we need this functionality for other external systems? I think it might be useful to have some kind of tracing/correlation mechanism in our logger implementation. Maybe a follow-up ticket? What do you think?

Approved with the changes/clarifications regarding names above.

src/modules/itslearning/itslearning.service.ts
```diff
@@ -34,9 +35,10 @@ export class ItsLearningIMSESService {
 
     public async send<ResponseBody, ResultType>(
         action: IMSESAction<ResponseBody, ResultType> | IMSESMassAction<ResponseBody, ResultType>,
+        syncId?: string,
```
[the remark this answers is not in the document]
marode-cap: itslearning calls them "syncID"s, so I've used their language here (since this is specifically for itslearning)

src/modules/itslearning/event-handlers/itslearning-persons.event-handler.ts
marode-cap: I feel like this is more of a specific for the event-handler. The event itself has an ID, but every event might send multiple requests to itslearning. Because of this, I concatenate this here.

marode-cap:
> Do we need this functionality for other external systems? I think it might be useful to have some kind of tracing/correlation mechanism in our logger implementation. Maybe a follow-up ticket? What do you think?
> 
> Approved with the changes/clarifications regarding names above.

Kristoff and I already talked and the goal would be to have per-request correlation IDs. When that comes to fruition, we can just use that ID in the event instead of generating it.

PR #844, opened 2024-12-16
Spsh-1574: Fortschrittsanzeige

phaelcg:
# Description
- Implemented create many for the import-data-repository
- Added totalDataItemImported to the importvorgang-entity
- refactored import-workflow 
- refactored import-event-handler to import each 25 data-items 
- Added totalDataItemImported to the response status
- Unit tests & integration tests

## Links to Tickets or other pull requests
https://ticketsystem.dbildungscloud.de/browse/SPSH-1574
https://github.com/dBildungsplattform/schulportal-client/pull/396

## Approval for review
- [ ] QA: In addition to review, the code has been manually tested (if manual testing is possible)
- [ ] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

PR #859, opened 2024-12-20
SPSH-1654: Paginierung des Importergebnis

phaelcg:
# Description
- Paginierung des Importergebnis mit Limit & Offset
- Neuer Endpunkt GET/api/import/importedUsers? für die Abfrage des Importergebnis als JSON
- Refactoring des Import-Workflows
- Anpassung der Logs für die Importstatus
- Fix für den Import
- Unit- & Integrationtests

https://ticketsystem.dbildungscloud.de/browse/SPSH-1654

## Approval for review


YoussefBouch:
#844

src/modules/import/domain/import-workflow.ts
```diff
@@ -296,43 +284,24 @@ export class ImportWorkflow {
...
+        const importVorgang: ImportVorgang<true> = importVorgangResult.value;
         if (importVorgang.status !== ImportStatus.FINISHED) {
             return {
                 ok: false,
                 error: new ImportDomainError('ImportVorgang is still in progress', importvorgangId),
             };
         }
 
-        this.initialize(importVorgang.organisationId, importVorgang.rolleId);
+        this.initialize(importVorgang.organisationId!, importVorgang.rolleId!);
```
marode-cap, line 304: Also null-assertions here, to me this looks like the underlying data-structure might need a rework.
If these fields can actually be undefined, there should probably be a check for it

src/modules/import/domain/import-workflow.ts
```diff
@@ -458,7 +483,8 @@ export class ImportWorkflow {
             return new ImportCSVFileEmptyError();
         }
 
-        if ((csvContent.match(/[\r\n]/g) || []).length - 1 > this.CSV_MAX_NUMBER_OF_USERS) {
+        const numberOfRows: number = Math.ceil((csvContent.match(/[\r\n]/g) || []).length / 2);
+        if (numberOfRows > this.CSV_MAX_NUMBER_OF_USERS) {
```
marode-cap, line 487: Why the divide by 2?

marode-cap (changes requested):

src/modules/import/domain/import-workflow.ts
YoussefBouch: There is apparently a check for it in the same file:

image

src/modules/import/api/imported-user.response.ts
```diff
@@ -0,0 +1,32 @@
...
+    @ApiProperty({ enum: ImportDataItemStatus, enumName: 'ImportDataItemStatus' })
+    public status: ImportDataItemStatus;
+
+    public constructor(importedDataItem: ImportDataItem<true>) {
+        this.klasse = importedDataItem.klasse!;
+        this.vorname = importedDataItem.vorname;
+        this.nachname = importedDataItem.nachname;
+        this.benutzername = importedDataItem.username!;
+        this.startpasswort = importedDataItem.password!;
+        this.status = importedDataItem.status!;
```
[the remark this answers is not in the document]
YoussefBouch: I changed the status to be always present but things like benutzername (referrer) or startpasswort come from another object definition and it could indeed be undefined according to typescript... just won't happen in a real life scenario..

src/modules/import/domain/import-workflow.ts
marode-cap: Ah, I missed that the regex contained a character group. We could switch the regex to `/(\n|\r\n)/g` which should encompass all the different line-endings used by Unix/Windows

src/modules/import/domain/import-workflow.ts
marode-cap: Ah, but because it is at a different location TS can't infer the asserted type. Could you at least add a comment when using null-assertions, so the reader of the code knows where the check is?

YoussefBouch: done

src/modules/import/domain/import-workflow.ts
pkleybolte: I don't see a reason to divide by 2. There should be one occurence of \r\n per line.

YoussefBouch: I used: `(/\r?\n/g)` it should look for \n everytime and an \r optionally

src/modules/import/domain/import-workflow.ts
pkleybolte: Same file, but not the same method. It is not likely, but they could be null here.

src/modules/import/domain/import-workflow.ts
YoussefBouch: > I don't see a reason to divide by 2. There should be one occurence of \r\n per line.

Right but the \r and \n are counted here individually and so they give apparently double the correct number of rows. The new (/\r?\n/g) should solve it

marode-cap:
Tests are still failing

PR #869, opened 2025-01-10
Dbp 1132 health ingress

mcpovel:
Added ingress to allow access to /api/health with basic authentication.


charts/dbildungs-iam-server/templates/_dbildungs-iam-server-envs.tpl
```diff
@@ -126,4 +126,9 @@
     secretKeyRef:
         name: {{ default .Values.auth.existingSecret .Values.auth.secretName }}
         key: vidis-keycloak-role
+- name: BASIC_AUTH
```
aimee-889, line 129: i wuold prefer if a METRICS_ is placed before it so no one gets confused

charts/dbildungs-iam-server/templates/backend-ingress.yaml
```diff
@@ -30,4 +30,41 @@ spec:
...
+  name: {{ template "common.names.name" . }}-health-backend
+  namespace: {{ template "common.names.namespace" . }}
+  labels:
+    {{- include "common.labels" . | nindent 4 }}
+  annotations:
+    nginx.ingress.kubernetes.io/rewrite-target: /health
+    # authentication type
+    nginx.ingress.kubernetes.io/auth-type: basic
+    # name of the secret that contains the user/password definitions
+    nginx.ingress.kubernetes.io/auth-secret: {{ .Values.auth.secretName }}
```
aimee-889, line 48: i think the comments in 45 an 47 are not adding much additional value. It is easiest to use the existing dbildungs-iam-server secret. The ingress also only seems to use the "auth" key so no other secrets would ever get exposed, still i am not sure whether creating a extra secret would the cleaner way. Has anyone an opinion on that?

charts/dbildungs-iam-server/templates/_dbildungs-iam-server-envs.tpl
mcpovel: Changed as suggested.

PR #876, opened 2025-01-16
Spsh 1495

tikle-cap:
https://ticketsystem.dbildungscloud.de/browse/SPSH-1495

### DIFF
--- a/src/modules/person/domain/person-time-limit-info.service.ts
+++ b/src/modules/person/domain/person-time-limit-info.service.ts
@@ -68,8 +68,25 @@ export default class PersonTimeLimitService {
         );
 
         const organisationInfos: (PersonTimeLimitInfo | null)[] = await Promise.all(organisationPromises);
-        lockInfos.push(...organisationInfos.filter((info: PersonTimeLimitInfo | null) => info !== null));
+        const validInfos: PersonTimeLimitInfo[] = organisationInfos.filter(
+            (info: PersonTimeLimitInfo | null): info is PersonTimeLimitInfo => info !== null,
+        );
 
+        const latestInfosMap: Map<string, PersonTimeLimitInfo> = new Map();
+
+        validInfos.forEach((info: PersonTimeLimitInfo) => {
+            const orgName: string | undefined = info.school;
+            if (!orgName) {
+                return;
+            }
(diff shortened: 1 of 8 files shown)


src/modules/person/domain/person-time-limit-info.service.ts
```diff
@@ -41,6 +45,36 @@ export default class PersonTimeLimitService {
             lockInfos.push(new PersonTimeLimitInfo(TimeLimitOccasion.NO_KONTEXTE, noKontexteDeadline));
         }
 
-        return lockInfos;
+        const personenKontexte: Personenkontext<true>[] = await this.dBiamPersonenkontextRepo.findByPerson(personId);
+
+        const timeLimitInfosPromises: Promise<PersonTimeLimitInfo | null>[] = personenKontexte.map(
+            async (personenKontext: Personenkontext<true>) => {
+                if (personenKontext.befristung != null) {
+                    const personenKontextExpires: Date = new Date(personenKontext.befristung);
```
marode-cap, line 53: Refrain from using loose comparisons. Also comparing against `null` can miss `undefined`
```suggestion
            async (personenKontext: Personenkontext<true>) => {
                if (!!personenKontext.befristung) {
                    const personenKontextExpires: Date = new Date(personenKontext.befristung);
```

src/modules/person/domain/person-time-limit-info.service.ts
```diff
@@ -41,6 +45,36 @@ export default class PersonTimeLimitService {
...
+                    }
+                }
+                return null;
+            },
+        );
+
+        const timeLimitInfos: (PersonTimeLimitInfo | null)[] = await Promise.all(timeLimitInfosPromises);
+        const validTimeLimitInfos: PersonTimeLimitInfo[] = timeLimitInfos.filter(
+            (info: PersonTimeLimitInfo | null): info is PersonTimeLimitInfo => info !== null,
+        );
```
marode-cap, line 77: Less of a problem here because we can be sure `info` is not `undefined`, but I'd still prefer this
```suggestion
        const validTimeLimitInfos: PersonTimeLimitInfo[] = timeLimitInfos.filter(
            (info: PersonTimeLimitInfo | null): info is PersonTimeLimitInfo => !!info,
        );
```

marode-cap (approved):

PR #886, opened 2025-01-27
DBP-1271-correct-service-selector

aimee-889:
# Description

We have seen that Updates lead to a Downtime when logged in. This is due to the service being updated and requiring the new version to be in the pod. But the new pod is just starting up and the old ones do not match the new version and get ignored. Therefore we need to remove the line that also filters for version.

## Approval for review


aimee-889:
https://github.com/dBildungsplattform/dbildungs-iam-server/pull/887 reopened it

PR #889, opened 2025-01-29
Spsh 1766

tikle-cap:
https://ticketsystem.dbildungscloud.de/browse/SPSH-1766


DPDS93CT (approved):
endpoints not secured by this change will be handled in SPSH-1787

PR #892, opened 2025-02-04
SPSH-1704

YoussefBouch:
# Description

- The selected Rollen sent through the UI are not an array instead of just 1 ID.
- The same Rollen are sent back so the response was also modified to be an array of Rollen.
- The same checks will be applied to all the Rollen sent to evaluate if the operation can be commited before creating the person.
- Removed deprecated Endpoint and it's test.

## Approval for review


src/modules/personenkontext/domain/personenkontext-workflow.ts
```diff
@@ -178,15 +178,20 @@ export class PersonenkontextWorkflowAggregate {
...
+            for (const rolleId of this.selectedRolleIds) {
+                // eslint-disable-next-line no-await-in-loop
+                const referenceCheckError: Option<DomainError> = await this.checkReferences(
+                    this.selectedOrganisationId,
+                    rolleId,
+                );
+                if (referenceCheckError) {
+                    return referenceCheckError;
+                }
             }
```
marode-cap, line 192: Can we replace this loop with a Promise.all (or Promise.allSettled or similar)?

src/modules/personenkontext/api/response/dbiam-personenkontext-workflow-response.ts
```diff
@@ -23,12 +25,15 @@ export class PersonenkontextWorkflowResponse {
     })
     public readonly selectedOrganisation?: string;
 
+    @ArrayUnique()
+    @IsUUID(undefined, { each: true })
+    @TransformToArray()
```
marode-cap, line 30: Responses don't need validators (They won't do anything)

marode-cap (approved):

src/modules/personenkontext/api/response/dbiam-personenkontext-workflow-response.ts
YoussefBouch: Done

src/modules/personenkontext/domain/personenkontext-workflow.ts
YoussefBouch: Done

PR #938, opened 2025-04-24
SPSH-2075: Add ability to sync MailAlternativeAddress via LDAP-Sync

DPDS93CT:
# Description

- add ability to sync MailAlternativeAddress via LDAP-Sync

- https://github.com/dBildungsplattform/dbildungs-iam-server/pulls
- https://ticketsystem.dbildungscloud.de/browse/SPSH-2075

## Approval for review


src/modules/authentication/services/session-access-token.middleware.ts
```diff
@@ -58,7 +58,8 @@ export class SessionAccessTokenMiddleware implements NestMiddleware {
                     }
                 } else {
                     req.logout((err: unknown) => {
-                        this.logger.logUnknownAsError('Logout Failed', err);
+                        //using logUnknownAsError won't log anymore details but a warning that err is UNDEFINED
+                        this.logger.error('Logout Failed', err);
```
marode-cap, line 62: This isn't really part of the ticket, but do we know that `err` will always be undefined? If it is, we don't need the error anyway. If it isn't, it might make more sense to update the `logUnknownAsError` so it can handle `undefined` without logging warning

marode-cap (approved):

PR #1019, opened 2025-07-31
SPSH-2343: add merkmale to service-providers

clauyan:
# Description

## Approval for review


src/modules/rolle/specification/only-assignable-sps.ts
```diff
@@ -0,0 +1,32 @@
+import { xor } from 'lodash-es';
+import { ServiceProvider } from '../../service-provider/domain/service-provider.js';
+import { CompositeSpecification } from '../../specification/specifications.js';
+
+import { ServiceProviderMerkmal } from '../../service-provider/domain/service-provider.enum.js';
+import { ServiceProviderRepo } from '../../service-provider/repo/service-provider.repo.js';
+import { Rolle } from '../domain/rolle.js';
+export class NurNachtraeglichZuweisbareServiceProvider extends CompositeSpecification<Rolle<boolean>> {
```
YoussefBouch, line 8: only-assignable-service-providers.specification.ts is a clearer name imo, what do you think?

YoussefBouch (approved):

src/modules/rolle/specification/only-assignable-sps.ts
```diff
@@ -0,0 +1,32 @@
...
+        private readonly oldRolle: Rolle<true>,
+    ) {
+        super();
+    }
+
+    public async isSatisfiedBy(updatedRolle: Rolle<boolean>): Promise<boolean> {
+        const changedServiceProviderIds: string[] = xor(
+            this.oldRolle.serviceProviderIds,
+            updatedRolle.serviceProviderIds,
+        );
```
YoussefBouch, line 20:
```suggestion
        const changedServiceProviderIds: string[] = xor(
            this.oldRolle.serviceProviderIds,
            updatedRolle.serviceProviderIds,
        );
        
          if (changedServiceProviderIds.length === 0) {
          // nothing changed, so the constraint is trivially satisfied
          return true;
           }
```
Early return if nothing was changed (Since its possible to make send a request through the UI with no change at all)

PR #1053, opened 2025-09-10
DBP-1587 Kafka Connectivity Backend

casparneumann-cap:
# Description

## Approval for review


casparneumann-cap:
> Also since it is disabled Kafka is set to false for the dbildungs-iam-server-db-migration-apply and dbildungs-iam-server-keycloak-data-migration init container as well as the seeding init container @casparneumann-cap i think kafka then should be set o true aas well? @casparneumann-cap . If so i think this line has to go and it might be that some Helm Charts changes/additions become necessary when kafka is enabled in these containers too: https://github.com/dBildungsplattform/dbildungs-iam-server/blob/main/src/console/main.ts#L6

We’ve intentionally disabled Kafka for the dbildungs-iam-server-db-migration-apply, dbildungs-iam-server-keycloak-data-migration, and the seeding init containers for now. This decision was based on previous issues we encountered when Kafka was enabled during these initialization phases. Our plan is to revisit this setup and enable Kafka at a later stage during init (when removing legacy event system)

aimee-889:
Okay, so i get that it should stay disabled for now. Then just please keep in mind, that some helm charts adjustment may (not sure) get needed once you decide to turn it back in again

PR #1093, opened 2025-10-06
SPSH-2647: Endpunkte für Ergebnisliste und Details für Angebote

clauyan:
# Description

## Approval for review

### DIFF
--- a/src/modules/service-provider/api/organisation-ref.response.spec.ts
+++ b/src/modules/service-provider/api/organisation-ref.response.spec.ts
@@ -1,4 +1,3 @@
-
 import { DoFactory } from '../../../../test/utils';
 import { Organisation } from '../../organisation/domain/organisation.js';
 import { OrganisationRefResponse } from './organisation-ref.response.js';


src/modules/service-provider/api/provider.controller.ts
```diff
@@ -115,4 +122,82 @@ export class ProviderController {
...
+                serviceProviders,
+            );
+
+        return serviceProvidersWithRollenAndErweiterungen.map(
+            (spWithData: ManageableServiceProviderWithLinkedObjects) =>
+                new ManageableServiceProviderListEntryResponse(
+                    spWithData.serviceProvider,
+                    spWithData.organisation,
+                    spWithData.rollen,
+                    spWithData.rollenerweiterungen,
```
YoussefBouch, line 155: I don't think we need to return the rollenerweiterungen in this case, it's just the Ids anyways so probably and not displayable. Can you confirm?

src/modules/service-provider/domain/service-provider.service.ts
```diff
@@ -19,6 +20,28 @@ import { ServiceProviderRepo } from '../repo/service-provider.repo.js';
...
+    };
+    rolle: {
+        id: RolleID;
+        name: string;
+    };
+    serviceProvider: {
+        id: ServiceProviderID;
+        name: string;
+    };
+};
```
YoussefBouch, line 44: I'd maybe put these in a separate `service-provider.types.ts` and use them wherever needed (in this case here and in the controller) to keep the service class compact.

src/modules/service-provider/domain/service-provider.service.ts
```diff
@@ -81,6 +104,58 @@ export class ServiceProviderService {
...
+            rolle: {
+                id: rollenerweiterung.rolleId,
+                name: rollen.get(rollenerweiterung.rolleId)?.name ?? '',
+            },
+            serviceProvider: {
+                id: rollenerweiterung.serviceProviderId,
+                name: serviceProvider.get(rollenerweiterung.serviceProviderId)?.name ?? '',
+            },
+        }));
+    }
```
YoussefBouch, line 157: Reduces redundant lookups:

```suggestion
return rollenerweiterungen.map(
  (rollenerweiterung): RollenerweiterungForManageableServiceProvider => {
    const org: Organisation<true> = organisationen.get(rollenerweiterung.organisationId);
    const rolle: Rolle<true>= rollen.get(rollenerweiterung.rolleId);
    const sp: Serviceprovider<true> = serviceProvider.get(rollenerweiterung.serviceProviderId);

    return {
      organisation: {
        id: rollenerweiterung.organisationId,
        name: org?.name ?? '',
        kennung: org?.kennung,
      },
      rolle: {
        id: rollenerweiterung.rolleId,
        name: role?.name ?? '',
      },
      serviceProvider: {
        id: rollenerweiterung.serviceProviderId,
        name: sp?.name ?? '',
      },
    };
  },
);

```

YoussefBouch (changes requested):

PR #1194, opened 2026-04-01
SPSH-3353

martkrau:
# Description

## Approval for review


src/modules/email-microservice/domain/email-resolver.service.ts
```diff
@@ -51,6 +52,41 @@ export class EmailResolverService {
...
+                    const status: EmailAddressStatus = this.mapStatus(emailResponse.status);
+                    result.set(personId, new PersonEmailResponse(status, emailResponse.address));
+                } else {
+                    result.set(personId, undefined);
+                }
+            }
+            return result;
+        } catch (error) {
+            this.logger.logUnknownAsError(`Failed to fetch emails for persons`, error);
+            return new Map<PersonID, PersonEmailResponse | undefined>();
```
marode-cap: Don't just return a result if something went wrong. The caller needs to know that fetching the e-mails failed

src/email/modules/core/api/dtos/params/find-email-addresses-by-spsh-person-ids.bodyparams.ts
```diff
@@ -0,0 +1,16 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { ArrayMinSize, IsArray, IsNotEmpty, IsString, IsUUID } from 'class-validator';
+
+export class FindEmailAddressBySpshPersonIdsBodyParams {
+    @IsArray()
+    @ArrayMinSize(2500)
```
marode-cap:
```suggestion
    @ArrayMaxSize(2500)
```
We might even be able to raise this limit, the ticket states that the endpoint should be able to *handle* 2500 IDs, not that it can only be called with at least 2500 IDs

src/email/modules/core/api/dtos/params/find-email-addresses-by-spsh-person-ids.bodyparams.ts
```diff
@@ -0,0 +1,16 @@
...
+    @IsString({ each: true })
```
marode-cap: `IsString` is unnecessary when `IsUUID` exists. same with `IsNotEmpty`

src/email/modules/core/api/controller/email-read.controller.ts
```diff
@@ -97,4 +98,43 @@ export class EmailReadController {
 
         return new EmailAddressResponse(emailAddress, latestStatus, this.oxService.contextID);
     }
+
+    @Post('spshpersons')
+    @Public()
+    @ApiOperation({ description: 'Get email-addresses by personIds.' })
+    @ApiOkResponse({
+        description: 'The email-addresses for corresponding persons were successfully returned.',
+        type: [EmailAddressResponse],
```
marode-cap, line 107: The type is incorrect.
Nestjs Swagger doesn't really support top-level maps/dictionaries, but I think we can get close enough with this so the generated api-docs make more sense

```suggestion
        schema: {
            type: 'object',
            additionalProperties: { $ref: getSchemaPath(EmailAddressResponse), nullable: true },
        ],
```

src/email/modules/core/api/controller/email-read.controller.ts
```diff
@@ -97,4 +98,43 @@ export class EmailReadController {
...
+    })
+    @ApiInternalServerErrorResponse({
+        description: 'Internal server error while getting email-addresses by personIds.',
+    })
+    public async findEmailAddressesByPersonIds(
+        @Body() findEmailAddressBySpshPersonIdsBodyParams: FindEmailAddressBySpshPersonIdsBodyParams,
+    ): Promise<Record<string, EmailAddressResponse | null>> {
+        const personIds: string[] = Array.from(new Set(findEmailAddressBySpshPersonIdsBodyParams.spshPersonIds));
+        this.logger.info(`PersonIds: ${personIds.join(', ')}`);
```
marode-cap: I don't know if this is a good idea, this will get pretty long (like 100kb long if 2500 IDs are sent).
Maybe just logging the count (and possibly the first 10 IDs) is enough, clarify in daily

src/email/modules/core/persistence/email-address.repo.ts
```diff
@@ -112,6 +112,30 @@ export class EmailAddressRepo {
...
+            result.set(pid, null);
+        }
+
+        for (const entity of emailAddressEntities) {
+            const pid: string = entity.spshPersonId;
+
+            if (result.get(pid) === null) {
+                result.set(pid, mapEntityToAggregate(entity));
+            }
+        }
```
marode-cap: I'd much rather combine these two loops and only do one `.set` on the map, to save on lookups.
Also, is there a need to even set an explicit null? Couldn't we just omit users if they have no e-mail?

src/modules/email-microservice/domain/email-resolver.service.ts
```diff
@@ -51,6 +52,41 @@ export class EmailResolverService {
         }
     }
 
+    public async findEmailsBySpshPersons(personIds: string[]): Promise<Map<PersonID, PersonEmailResponse | undefined>> {
```
marode-cap:
```suggestion
    public async findEmailsBySpshPersons(personIds: PersonID[]): Promise<Map<PersonID, PersonEmailResponse | undefined>> {
```

src/modules/email-microservice/domain/email-resolver.service.ts
```diff
@@ -51,6 +52,41 @@ export class EmailResolverService {
...
+            >();
+            for (const personId of personIds) {
+                const emailResponse: EmailAddressResponse | null | undefined = response.data[personId];
+                if (emailResponse) {
+                    const status: EmailAddressStatus = this.mapStatus(emailResponse.status);
+                    result.set(personId, new PersonEmailResponse(status, emailResponse.address));
+                } else {
+                    result.set(personId, undefined);
+                }
+            }
```
marode-cap: The more I look at this, the less I like it.
When requesting the emails the entire chain will be:
- Microservice loads from Database and builds a `Map`
- Just to immediately make the map into a plain object to send the response
- Server receives the plain object and immediately converts it back into a `Map`
This seems pretty excessive, is there a reason why we can't skip all the Maps and just work with objects? And maybe not even put an undefined/null into that object for persons without an email, that will also reduce memory usage.

We should probably talk with one of the architects about this. 
In my opinion these maps, at least the one in the microservice, are just wasting resources

src/email/modules/core/api/controller/email-read.controller.spec.ts
```diff
@@ -215,4 +216,135 @@ describe('EmailReadController', () => {
             expect(result).toEqual([]);
         });
     });
+
+    describe('findEmailAddressesByPersonIds', () => {
+        it('should return a map of personIds to EmailAddressResponse', async () => {
+            const spshPerson1Id: string = faker.string.uuid();
+            const params1: FindEmailAddressBySpshPersonIdPathParams = new FindEmailAddressBySpshPersonIdPathParams();
+            Object.assign(params1, { spshPerson1Id });
```
marode-cap: This is not used unless I'm missing something

src/email/modules/core/api/controller/email-read.controller.spec.ts
```diff
@@ -215,4 +216,135 @@ describe('EmailReadController', () => {
...
+                    {
+                        id: faker.string.uuid(),
+                        status: EmailAddressStatusEnum.ACTIVE,
+                    },
+                ],
+            });
+
+            const spshPerson2Id: string = faker.string.uuid();
+            const params2: FindEmailAddressBySpshPersonIdPathParams = new FindEmailAddressBySpshPersonIdPathParams();
+            Object.assign(params2, { spshPerson2Id });
```
marode-cap: This as well

marode-cap (changes requested):

src/email/modules/core/persistence/email-address.repo.ts
martkrau: We want to have every user in the result list, not just the ones with an email.

src/modules/email-microservice/domain/email-resolver.service.ts
martkrau: For now I removed the for loop and used just map()

src/email/modules/core/api/controller/email-read.controller.ts
```diff
@@ -97,4 +99,43 @@ export class EmailReadController {
...
+    @Post('spshpersons')
+    @Public()
+    @ApiOperation({ description: 'Get email-addresses by personIds.' })
+    @ApiOkResponse({
+        description: 'The email-addresses for corresponding persons were successfully returned.',
+        schema: {
+            type: 'object',
+            additionalProperties: { $ref: getSchemaPath(EmailAddressResponse), nullable: true },
+        },
+    })
```
marode-cap, line 108: Response was changed, so this should now use `type: [EmailAddressResponse]`

src/email/modules/core/persistence/email-address.repo.ts
```diff
@@ -112,6 +112,30 @@ export class EmailAddressRepo {
         return emailAddressEntities.map(mapEntityToAggregate);
     }
 
+    public async findPrimaryBySpshPersonIds(spshPersonIds: string[]): Promise<EmailAddress<true>[]> {
+        const emailAddressEntities: Option<EmailAddrEntity[]> = await this.em.find(
+            EmailAddrEntity,
+            {
+                spshPersonId: { $in: spshPersonIds },
+                priority: 0,
+                statuses: { status: EmailAddressStatusEnum.ACTIVE },
```
marode-cap: Filtering status like this is not going to work, since this is a collection and we only want the newest one to matter. (This code will return all mails that were 'ACTIVE' at any point in time, so also currently 'SUSPENDED' mails)

Probably better to filter status in code (I don't think we can do this in the database with MikroORM)

src/email/modules/core/persistence/email-address.repo.ts
```diff
@@ -112,6 +112,30 @@ export class EmailAddressRepo {
...
+
+        const result: EmailAddress<true>[] = [];
+        const seen: Set<string> = new Set<string>();
+
+        for (const entity of emailAddressEntities) {
+            if (!seen.has(entity.spshPersonId)) {
+                seen.add(entity.spshPersonId);
+                result.push(mapEntityToAggregate(entity));
+            }
+        }
```
marode-cap, line 129: There can only ever be one mail with priority 0 (per person), no need to check if we've already seen that personID (just loop and check the newest status)

src/email/modules/core/persistence/email-address.repo.ts
```diff
@@ -112,6 +112,30 @@ export class EmailAddressRepo {
...
 
+    public async findPrimaryBySpshPersonIds(spshPersonIds: string[]): Promise<EmailAddress<true>[]> {
+        const emailAddressEntities: Option<EmailAddrEntity[]> = await this.em.find(
+            EmailAddrEntity,
+            {
+                spshPersonId: { $in: spshPersonIds },
+                priority: 0,
+                statuses: { status: EmailAddressStatusEnum.ACTIVE },
+            },
+            { orderBy: { spshPersonId: 'asc', priority: 'asc' }, populate: ['statuses'] },
```
marode-cap: We don't need ordering for this, so no need to do it and waste performance.
The populate is also not needed (the statuses-collection is marked as "eager" in the entity, so it will always be populated)

src/email/modules/core/persistence/email-address.repo.integration-spec.ts
```diff
@@ -173,6 +173,52 @@ describe('EmailRepo', () => {
...
+            const mail1: EmailAddress<true> = await createAndSaveMail(undefined, 1, spshPersonIds[0]);
+            await setStatus(mail1, EmailAddressStatusEnum.SUSPENDED);
+            const mail2: EmailAddress<true> = await createAndSaveMail(undefined, 0, spshPersonIds[0]);
+            await setStatus(mail2, EmailAddressStatusEnum.ACTIVE);
+            const mail3: EmailAddress<true> = await createAndSaveMail(undefined, 1, spshPersonIds[1]);
+            await setStatus(mail3, EmailAddressStatusEnum.SUSPENDED);
+            const mail4: EmailAddress<true> = await createAndSaveMail(undefined, 2, spshPersonIds[1]);
+            await setStatus(mail4, EmailAddressStatusEnum.SUSPENDED);
+            const mail5: EmailAddress<true> = await createAndSaveMail(undefined, 0, spshPersonIds[2]);
+            await setStatus(mail5, EmailAddressStatusEnum.ACTIVE);
```
marode-cap: Add a case where an E-Mail is Prio-0, with an old status of ACTIVE and a newer status of SUSPENDED

src/email/modules/core/persistence/email-address.repo.integration-spec.ts
```diff
@@ -173,6 +173,52 @@ describe('EmailRepo', () => {
...
+        });
+
+        it('should return primary email addresses for the given spshPersonIds', async () => {
+            const result: EmailAddress<true>[] = await sut.findPrimaryBySpshPersonIds(spshPersonIds);
+            expect(result).toHaveLength(2);
+            expect(result.find((e: EmailAddress<true>) => e.spshPersonId === spshPersonIds[0])!.priority).toBe(0);
+            expect(result.find((e: EmailAddress<true>) => e.spshPersonId === spshPersonIds[2])!.priority).toBe(0);
```
marode-cap: Could these be something like this for readability?
`expect(result).toEqual(expect.arrayContaining([mail1, mail2]))`

src/modules/schulconnex/domain/personeninfo/personeninfo.service.ts
```diff
@@ -54,7 +57,7 @@ export class PersonenInfoService {
         }
 
         if (permittedServiceProviderIds.size === 0) {
-            return [];
+            return { ok: true, value: [] };
```
marode-cap: Just going to write this once, but you've done this everywhere in these changes:
We have utility-methods to create results (`Ok(xxx)` and `Err(xxx)`)

Additionally there is also `expectOkResult(x)` and `expectErrResult(x)` for use in tests, so you don't need to do stuff like 
```ts
if (!res.ok) {
    throw new Error();
}```

File: src/modules/schulconnex/domain/personeninfo/personeninfo.service.ts [hunk 18]
```diff
@@ -88,15 +96,20 @@ export class PersonenInfoService {
             this.userLockRepo.findByPersonIds(personIds),
         ]);
 
+        if (!emailsForPersons.ok) {
+            return { ok: false, error: emailsForPersons.error };
+        }
```
marode-cap, line 102: Results can be directly returned if you want to pass along the error
```suggestion
        if (!emailsForPersons.ok) {
            return emailsForPersons;
        }
```

marode-cap (changes requested):

marode-cap (approved):

PR #1209, opened 2026-05-05
SPSH-3352: fix rolle deletion by implementing Rolle deletion service and refactoring related components

clauyan:
# Description

## Approval for review


src/modules/rolle/repo/rolle.repo.ts
```diff
@@ -486,32 +486,27 @@ export class RolleRepo {
...
         try {
-            //Cascade removal
-            await this.em.removeAndFlush(rolleEntity);
+            const entity: RolleEntity = this.em.create(RolleEntity, mapRolleAggregateToData(rolle.value));
+            await this.em.removeAndFlush(entity);
         } catch (ex) {
             if (ex instanceof ForeignKeyConstraintViolationException) {
                 return new RolleHatPersonenkontexteError();
             }
+            throw ex;
```
YoussefBouch, line 491: does it make more sense to throw errors from the controller as usual? this is the only method throwing from the repo

clauyan: In general, yes, but we don't have an established mechanism for that. The other methods just don't handle exceptions from the driver at all, so they also throw, but implicitly. I had a brief chat with @marode-cap and we agreed that it is better to rethrow the exception here instead of swallowing the error like before. The error is then filtered globally and mapped to the generic DB-Error.

YoussefBouch (approved):

PR #1225, opened 2026-05-27
SPSH-3034-CSRF

YoussefBouch:
# Description

## Approval for review


src/modules/authentication/api/csrf-token.guard.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        // Skip JWT requests (Authorization: Bearer ...)
+        if (this.isJwtRequest(request)) {
+            return true;
+        }
+
+        // Skip public routes
+        if (this.isPublicRoute(request.path)) {
+            return true;
+        }
```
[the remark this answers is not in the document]
YoussefBouch: @casparneumann-cap What do you think? Does Keycloak need these endpoints as of now?

If yes then I'd implement this suggestion.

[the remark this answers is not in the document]
casparneumann-cap: VERY IMPORTANT!: api-key communication is only internal communication we should skip CSRF validation when an api-key header is present. Otherwise Email Microservice Communication as Well as Keycloak Communication will be blocked.

src/modules/authentication/api/authentication.controller.ts
```diff
@@ -114,6 +118,42 @@ export class AuthenticationController {
...
+            const token: string = request.session?.csrfToken || this.csrfTokenService.generateToken(request);
+
+            this.logger.info('CSRF token provided to authenticated user');
+
+            return new CsrfTokenResponse(token);
+        } catch (error) {
+            this.logger.error('Failed to generate CSRF token', error);
+            return new CsrfTokenErrorResponse('Failed to generate CSRF token');
+        }
+    }
```
casparneumann-cap, line 165: This CSRF-token endpoint advertises 401 for unauthenticated callers, but the implementation just returns an error object, so Nest will still send 200 for both auth failures and token-generation failures.

casparneumann-cap (changes requested):

src/modules/authentication/api/authentication.controller.ts
YoussefBouch: Used UnauthorizedException for 401 and InternalServerErrorException for 500.

src/modules/authentication/api/csrf-token.guard.ts
```diff
@@ -0,0 +1,50 @@
...
+    private isJwtRequest(request: Request): boolean {
+        const authHeader: string = request.headers.authorization || '';
+        return authHeader.startsWith('Bearer ');
+    }
+
+    private isPublicRoute(path: string): boolean {
+        const publicRoutes: string[] = ['/api/health', '/api/auth/login', '/api/docs', '/api/docs-json'];
+        return publicRoutes.some((route: string) => path.startsWith(route));
+    }
+}
```
casparneumann-cap: I think this hard-coded list is very clunky. We’ll have to keep it in mind going forward. Also, e.g. the internal email cron controller from the new microservice is missing, for example. Isn’t there a way to automatically retrieve this information via the public annotation, perhaps with some customization?

src/modules/authentication/services/csrf-token.service.ts
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        // Query fallback
+        if (request.query['csrfToken']) {
+            return request.query['csrfToken'] as string;
+        }
+
+        return undefined;
+    }
+}
```
casparneumann-cap, line 52: Instead of  "as string" I suggest working with unknown + typeguard.

casparneumann-cap (changes requested):

casparneumann-cap (approved):

PR #1230, opened 2026-06-03
SPSH-3669

casparneumann-cap:
# Description

## Approval for review


src/modules/service-provider/repo/service-provider.repo.ts
```diff
@@ -151,6 +151,27 @@ export class ServiceProviderRepo {
...
+            ServiceProviderEntity,
+            {
+                providedOnSchulstrukturknoten: { $in: organisationIds },
+                vidisAngebotId: { $ne: null },
+            },
+            {
+                exclude,
+                populate: ['merkmale'],
+            },
+        );
```
marode-cap:
```suggestion
        const serviceProviders: ServiceProviderEntity[] = await this.em.find(
            ServiceProviderEntity,
            {
                providedOnSchulstrukturknoten: { $in: organisationIds },
                vidisAngebotId: { $ne: null },
            },
            {
                exclude: ['logo'] as const,
                populate: ['merkmale'],
            },
        );
```

src/modules/vidis/core/vidis.sync-service.ts
```diff
@@ -0,0 +1,376 @@
...
+        schoolOffset: number,
+        permissions: IPersonPermissions,
+    ): Promise<void> {
+        const [schools, total]: Counted<Organisation<true>> = await this.organisationRepo.findBy(
+            new OrganisationScope()
+                .findBy({
+                    typ: 'SCHULE',
+                })
+                .paged(schoolOffset, this.vidisConfig.SYNC_SCHOOLS_PAGE_SIZE),
+        );
```
marode-cap, line 100: Does the default scope contain a default ordering? Paging might not be stable otherwise.

charts/dbildungs-iam-server/values.yaml
```diff
@@ -357,11 +359,11 @@ cronjobs:
...
-#    cron-trigger-6:
-#      schedule: 30 0 * * *
-#      endpoint: '/api/cron/vidis-angebote'
-#      httpMethod: 'PUT'
-#      script: 'cron_trigger.sh'
+    cron-trigger-6:
+      schedule: "30 0 1 6 *"
+      endpoint: '/api/cron/vidis-sync'
+      httpMethod: 'PUT'
+      script: 'cron_trigger.sh'
```
marode-cap, line 366: Should this already be activated? Would affect prod after the next release

src/modules/vidis/core/vidis.sync-service.ts
```diff
@@ -0,0 +1,376 @@
...
+import { ConfigService } from '@nestjs/config';
+import { VidisApiAdapter } from '../adapter/domain/vidis-api.adapter.js';
+
+type VidisSchoolActivatedAngebot = {
+    angebot: VidisServiceResponseAngebot;
+    date: string;
+};
+
+type VidisAngeboteByOrganisationId = Record<string, VidisSchoolActivatedAngebot[]>;
+type VidisOrganisationIdByKennung = Record<string, string>;
```
marode-cap, line 36: Any reasons why this doesn't use a Map<>?

src/modules/vidis/core/vidis.sync-service.ts
```diff
@@ -0,0 +1,376 @@
...
+                    syncOperations.push(Promise.reject(deleteRollenerweiterungenResult.error));
+                }
+            } catch (error) {
+                const rejectionReason: Error = error instanceof Error ? error : new Error(String(error));
+                syncOperations.push(Promise.reject(rejectionReason));
+            }
+        }
+
+        return Promise.allSettled(syncOperations).then((results: PromiseSettledResult<unknown>[]) => {
+            const failedOperations: PromiseSettledResult<unknown>[] = results.filter(
```
marode-cap: why use `.then()` instead of awaiting the promise? This is just unnecessary indendation

src/modules/vidis/core/vidis.sync-service.ts
```diff
@@ -0,0 +1,376 @@
...
+                (angebotInDb: ServiceProvider<true>) =>
+                    angebotInDb.vidisAngebotId !== undefined && !vidisAngebotIds.has(angebotInDb.vidisAngebotId),
+            )
+            .map((angebotInDb: ServiceProvider<true>) => angebotInDb.id);
+
+        if (missingAngeboteInDb.length === 0 && serviceProviderIdsMissingInVidis.length === 0) {
+            this.logger.info(
+                `No differences between VIDIS API and database for school with organisationId: ${organisationId}`,
+            );
+            return Promise.resolve();
```
marode-cap: this is already an async function, `return;` should suffice

charts/dbildungs-iam-server/values.yaml
casparneumann-cap: It will not run in the next 12 Months due to "30 0 1 6 *".

src/modules/vidis/core/vidis.sync-service.ts
casparneumann-cap: A Simple Record is enough here. Since we don't use any Map features we are not losing any behavior we need.

src/modules/vidis/core/vidis.sync-service.ts
casparneumann-cap: I added .sortBy('id', ScopeOrder.ASC) to be sure

marode-cap (approved):

PR #1237, opened 2026-06-09
SPSH-3777

pkleybolte:
# Description

https://ticketsystem.dbildungscloud.de/browse/SPSH-3777

## Approval for review


src/modules/rolle/domain/rolle-find.service.ts
```diff
@@ -83,6 +83,85 @@ export class RolleFindService {
...
+        const [candidateRollen]: Counted<Rolle<true>> = await this.rolleRepo.findBy({
+            searchStr: params.searchStr,
+            allowedOrganisationIds,
+            rollenArten: params.rollenArten,
+        });
+
+        let allowedRollen: Rolle<true>[] = candidateRollen;
+        if (targetOrganisationIds !== undefined && targetOrganisationIds.length > 0) {
+            const organisationIdsWithParents: OrganisationID[] =
+                await this.getOrganisationIdsWithParents(targetOrganisationIds);
```
YoussefBouch: `getOrganisationIdsWithParents` is called for `allowedOrganisationIds` already, maybe you can re-use it.

src/modules/rolle/domain/rolle-find.service.ts
```diff
@@ -83,6 +83,85 @@ export class RolleFindService {
...
+                                    return canAssignResult.ok;
+                                }),
+                            )
+                        ).some(Boolean);
+
+                        return canBeAssignedToAnyTargetOrga ? rolle : null;
+                    }),
+                )
+            ).filter((rolle: Rolle<true> | null): rolle is Rolle<true> => rolle !== null);
+        }
```
YoussefBouch: All `candidateRollen` are evaluated against `targetOrganisationsWithParents` with no early exit when a match is found. I think we could early exit as soon as a match is found instead of evaluating against every other orga:

```
allowedRollen = (
    await Promise.all(candidateRollen.map(async (rolle) => {
        for (const organisation of targetOrganisationsWithParents) {
            const result = await rolle.canBeAssignedToOrga(organisation);
            if (result.ok) return rolle; // here we early exit
        }
        return null;
    }))
).filter((rolle): rolle is Rolle<true> => rolle !== null);
```

YoussefBouch (approved):

PR #1244, opened 2026-06-24
SPSH-2971

YoussefBouch:
# Description

## Approval for review


src/modules/service-provider/api/provider.controller.ts
clauyan, line 1: `findRollenerweiterungenByServiceProviderId` does permission-checks, which we generally avoid in controllers. Consider moving the logic related to fetching the rollenerweiterungen into a service, i.e. `rolle/domain/rollenerweiterung-find.service.ts`.

src/modules/service-provider/api/provider.controller.spec.ts
```diff
@@ -271,7 +278,7 @@ describe('Provider Controller Test', () => {
             expect(result.items[0]?.organisationKennung).toBe('FixedOrgaKennung2');
         });
 
-        it('should throw MissingPermissionsError when user lacks permission when filtering for orga', async () => {
+        it.only('should throw MissingPermissionsError when user lacks permission when filtering for orga', async () => {
```
clauyan: Leftover `only`

src/modules/rolle/repo/rollenerweiterung.repo.integration-spec.ts
clauyan, line 1: Consider adding an integration test for the new code path with `rolleIds`

src/modules/rolle/repo/rolle.repo.integration-spec.ts
clauyan, line 1: Add an integration test for `findRollenAuthorized` with `systemrechte`.

src/modules/rolle/api/find-rolle-query.param.ts
```diff
@@ -22,19 +23,35 @@ export class FindRolleQueryParams extends PagedQueryParams {
...
-    public readonly organisationId?: string;
+    public readonly organisationId?: OrganisationID;
 
     @IsOptional()
-    @IsEnum(RollenSystemRechtEnum)
+    @IsUUID(undefined, { each: true })
+    @TransformToArray()
+    @ArrayUnique()
+    @ApiProperty({
+        description: 'The ids of the selected Rollen',
```
clauyan: Add an explanation, that these ids have an or-relationship with the other parameters. Optionally add that this required by our implementation of Autocomplete in the frontend.

src/modules/rolle/api/rolle.controller.integration-spec.ts
clauyan, line 1: Consider adding integration tests for different combinations of `systemrechte` and `rolleIds`

clauyan (changes requested):

src/modules/service-provider/api/provider.controller.ts
YoussefBouch: Yeah, I agree 👍 it would be cleaner to move permissions check from the controllers into a separate service.

For this ticket though, I’d prefer to leave it as-is, otherwise it will likely grow quite a bit (refactoring + test adjustments) and we might go out of scope pretty quickly and have to postpone merging.
It’s also not something that was newly introduced, it’s been there for a while already. What do you think?

clauyan: Hm, I worry that it will just stay as is, if we don't do it, but I agree. We should prioritize delivering the feature on time.

src/modules/rolle/repo/rollenerweiterung.repo.ts
```diff
@@ -296,6 +297,10 @@ export class RollenerweiterungRepo {
             qb = qb.andWhere({ organisationId: { $in: organisationIds } });
         }
 
+        if (rolleIds && rolleIds.length > 0) {
+            qb = qb.andWhere({ rolleId: { $in: rolleIds } });
```
clauyan, line 301: This implementation returns only Rollenerweiterungen for specified `rolleIds`. That means the list in the frontend will show orga and ONLY the rollen, that are matched by the filter. Is that the intended behavior? Or should it be something like 'include the orga and all rollenerweiterungen, if a rollenerweiterung with this orga, rolle and angebot exists'?

YoussefBouch: According to one PO when an Admin sets a filter for a specific Rolle they shouldn't be seeing any other Rolle otherwise it's confusing (And I agree because in the case  you're describing the Rollenfilter is basically useless in combination with Schulfilter if you end up showing all Rollen for the Orga anyways so I'm pretty sure it wouldn't be a good thing to do it that way. 

But i'll bring it up again while testing it with them to make sure.

clauyan: Then it's fine imo

clauyan (approved):

PR #1248, opened 2026-07-03
SPSH-3820: add systemrecht MPT_ROLLEN_VERWALTEN and handling in findRollen

clauyan:
# Description

## Approval for review

### DIFF
--- a/src/modules/rolle/api/find-rolle-query.param.ts
+++ b/src/modules/rolle/api/find-rolle-query.param.ts
@@ -49,7 +49,7 @@ export class FindRolleQueryParams extends PagedQueryParams {
         required: false,
         isArray: true,
         description:
-            'The system right for which the roles should be available. Can only be ROLLEN_VERWALTEN, ROLLEN_ERWEITERN, MPT_ROLLEN_VERWALTEN or IMPORT_DURCHFUEHREN, or a combination of them.',
+            'The system right for which the roles should be available. Can only be ROLLEN_VERWALTEN, ROLLEN_ERWEITERN, MPT_ROLLEN_VERWALTEN or a combination of them or IMPORT_DURCHFUEHREN .',
     })
     @IsSystemrechtForRollenAdministration()
     public readonly systemrechte?: RollenSystemRechtEnum[];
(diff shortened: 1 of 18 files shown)


src/modules/rolle/api/find-rolle-query.param.ts
```diff
@@ -49,7 +49,7 @@ export class FindRolleQueryParams extends PagedQueryParams {
         required: false,
         isArray: true,
         description:
-            'The system right for which the roles should be available. Can only be ROLLEN_VERWALTEN, ROLLEN_ERWEITERN or both or IMPORT_DURCHFUEHREN.',
+            'The system right for which the roles should be available. Can only be ROLLEN_VERWALTEN, ROLLEN_ERWEITERN, MPT_ROLLEN_VERWALTEN or IMPORT_DURCHFUEHREN, or a combination of them.',
```
YoussefBouch:
```suggestion
            'The system right for which the roles should be available. Can only be ROLLEN_VERWALTEN, ROLLEN_ERWEITERN, MPT_ROLLEN_VERWALTEN or a combination of them or IMPORT_DURCHFUEHREN .',
```
Import SR is separate so it will never be a combination with it.

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -121,6 +121,20 @@ export class RolleController {
                 limit: queryParams.limit,
                 offset: queryParams.offset,
             });
+        } else if (
```
YoussefBouch, line 124: If im a Landesadmin in the global list of Roles then i'd send ROLLEN_VERWALTEN and MPT_ROLLEN_VERWALTEN which makes the array of length 2 and i'd expect to see the normal roles + MPT roles. so this might not work I guess?

clauyan: `RolleRepo.findRollenAuthorized` handles that case. See src/modules/rolle/repo/rolle.repo.ts:343.

src/modules/rolle/api/find-rolle-query.param.ts
clauyan: Added 4d25a6979880b923686052554fee052ff922d360

YoussefBouch (approved):

src/modules/rolle/api/rolle.controller.ts
YoussefBouch: Overlooked! LGTM :)

YoussefBouch:
Reminder: Change the target branch to SPSH-3997 ;)

PR #1250, opened 2026-07-10
SPSH-3819

casparneumann-cap:
# Description

## Approval for review


src/modules/service-provider/domain/service-provider-modification.service.ts
```diff
@@ -0,0 +1,265 @@
...
+            }
+        }
+
+        const hasVerfuegbarFuerRollenerweiterungMerkmalBeenRemoved: boolean =
+            !serviceProvider.merkmale.includes(ServiceProviderMerkmal.VERFUEGBAR_FUER_ROLLENERWEITERUNG) &&
+            existingProvider.merkmale.includes(ServiceProviderMerkmal.VERFUEGBAR_FUER_ROLLENERWEITERUNG);
+        if (hasRollenartenWhitelistChanged || hasVerfuegbarFuerRollenerweiterungMerkmalBeenRemoved) {
+            await this.rollenerweiterungRepo.deleteByServiceProviderIdAndRollenarten(
+                serviceProvider.id,
+                hasVerfuegbarFuerRollenerweiterungMerkmalBeenRemoved ? undefined : forbiddenRollenarten,
```
YoussefBouch: Relaxing a whitelist restriction (going from restricted => unrestricted) deletes every Rollenerweiterung row for that service provider, even ones for role types that were never forbidden and should be untouched. That's the opposite of what should happen. Widening a whitelist shouldn't delete anything I think.

[] (empty array) => should mean delete nothing, but the repo treats it like undefined and deletes everything. Can you check if this is intended or a bug?

A quick fix would be to guard the repo call so it's simply never invoked with an empty forbiddenRollenarten.

YoussefBouch (changes requested):

src/modules/service-provider/domain/service-provider-modification.service.ts
```diff
@@ -0,0 +1,265 @@
...
+            forbiddenRollenarten,
+        }: { hasRollenartenWhitelistChanged: boolean; forbiddenRollenarten: RollenArt[] } = this.getWhitelistChange(
+            existingProvider,
+            serviceProvider,
+        );
+
+        if (hasRollenartenWhitelistChanged && forbiddenRollenarten.length > 0) {
+            const attachedRollenError: Option<AttachedRollenError> = await this.getAttachedRollenError(
+                serviceProvider.id,
+                'Cannot update rollenartenWhitelist with conflicting rollenarten to attached to rollen',
```
YoussefBouch:
```suggestion
                'Cannot update rollenartenWhitelist with conflicting rollenarten  attached to rollen',
```

YoussefBouch (changes requested):

src/modules/service-provider/domain/service-provider-modification.service.ts
clauyan: Good catch, thanks! I added a test for regression, simplified the code and added the guard.

src/modules/service-provider/domain/service-provider-modification.service.ts
clauyan: Added

YoussefBouch (approved):

PR #1253, opened 2026-07-22
SPSH-3889

martkrau:
# Description

## Approval for review


src/modules/rolle/api/rolle.controller.ts
```diff
@@ -425,4 +433,90 @@ export class RolleController {
...
+    })
+    @ApiUnauthorizedResponse({ description: 'Not authorized to get RollenErweiterungen.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permission to get RollenErweiterungen.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while getting RollenErweiterungen.' })
+    public async findRollenerweiterungenForRolleAndOrga(
+        @Param() params: FindRolleByIdParams,
+        @Query() queryParams: FindRollenerweiterungQueryParams,
+        @Permissions() permissions: IPersonPermissions,
+    ): Promise<ServiceProviderResponse[]> {
+        const rolleResult: Result<Rolle<true>> = await this.rolleRepo.findByIdAuthorized(params.rolleId, permissions);
```
clauyan, line 453: `findByIdAuthorized` uses `ROLLEN_VERWALTEN` for its permission-check. I am not sure if we actually need to have `ROLLEN_VERWALTEN` or if `ROLLEN_ERWEITERN` should also be sufficient here. Since the ticket does not specify, I think we should clarify with an architect.

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -425,4 +433,90 @@ export class RolleController {
...
+        if (!rolleResult.ok) {
+            throw new EntityNotFoundError('Rolle', params.rolleId);
+        }
+        const permittedOrgaIds: PermittedOrgas = await permissions.getOrgIdsWithSystemrecht([
+            RollenSystemRecht.getByName(RollenSystemRechtEnum.ROLLEN_ERWEITERN),
+            RollenSystemRecht.getByName(RollenSystemRechtEnum.ROLLEN_VERWALTEN),
+        ]);
+
+        let rollenerweiterungen: Rollenerweiterung<true>[];
+        if (!queryParams.organisationId) {
```
clauyan, line 463: I think we should require `organisationId`. If I follow the ticket, the use case will be fetching angebote for a rolle at a specific organisation. I don't see the use case for getting all angebote via rollenerweiterungen *without* also knowing the organisation, but we risk stability, since the result could be very large.

That allows us to simplify this code a bit and we can avoid overfetching and filtering in memory.
check if requested orga is allowed -> fetch rollenerweiterungen for orga & rolle

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -425,4 +433,90 @@ export class RolleController {
...
+    @ApiOperation({ description: 'Apply Erweiterte Angebote changes for a rolle.' })
+    @ApiOkResponse({
+        description: 'The Erweiterten Angebote were successfully updated.',
+        type: ServiceProviderResponse,
+        isArray: true,
+    })
+    @ApiUnauthorizedResponse({ description: 'Not authorized to update RollenErweiterungen.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permission to update RollenErweiterungen.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while updating RollenErweiterungen.' })
+    public async applyRollenerweiterungChangesForRolle(
```
clauyan, line 495: I think this function should also live in `RollenerweiterungController`.

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -425,4 +433,90 @@ export class RolleController {
 
         return new RolleWithServiceProvidersResponse(rolle, rolleServiceProviders);
     }
+
+    @Get(':rolleId/rollenerweiterungen')
```
clauyan: I think the path and function name are misleading. Maybe something like `:rolleId/angeboteViaRollenerweiterungen`, `:rolleId/rollenerweiterungen/angebote` or even `:rolleId/organisation/:organisationId/rollenerweiterungen/angebote` is a bit clearer?

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -425,4 +433,90 @@ export class RolleController {
...
+    })
+    @ApiUnauthorizedResponse({ description: 'Not authorized to update RollenErweiterungen.' })
+    @ApiForbiddenResponse({ description: 'Insufficient permission to update RollenErweiterungen.' })
+    @ApiInternalServerErrorResponse({ description: 'Internal server error while updating RollenErweiterungen.' })
+    public async applyRollenerweiterungChangesForRolle(
+        @Param() params: ApplyRollenerweiterungForRollePathParams,
+        @Body() body: ApplyRollenerweiterungChangesBodyParams,
+        @Permissions() permissions: IPersonPermissions,
+    ): Promise<void> {
+        const rolleResult: Result<Rolle<true>> = await this.rolleRepo.findByIdAuthorized(params.rolleId, permissions);
```
clauyan, line 500: Same as above, this requires `ROLLEN_VERWALTEN`, but I think `ROLLEN_ERWEITERN` should also grant permission.

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
```diff
@@ -0,0 +1,242 @@
...
+};
+
+function isErrorResultForServiceProvider<T>(r: {
+    result: Result<T, DomainError>;
+}): r is TerrorResultForServiceProvider {
+    return r.result.ok === false;
+}
+
+@Injectable()
+export class ApplyRollenerweiterungWithRoleForAngeboteService {
```
clauyan: Since `Rolle` is a domain object, we should probably use the german spelling here and in the function name.

Did you check with an architect, if we really need two services (one for rolle, one for angebot) or if we can use a single `RollenerweiterungenModifyService` or similar?

src/modules/rolle/domain/apply-rollenerweiterungen-for-rolle-service.ts
```diff
@@ -0,0 +1,242 @@
...
+
+        const rollen: Map<string, Rolle<true>> = await this.rolleRepo.findByIds([rolleId]);
+        const rolle: Option<Rolle<true>> = rollen.get(rolleId);
+
+        if (!rolle) {
+            this.logger.error(
+                `applyRollenerweiterungChangesForRolle called by ${permissions.personFields.username} - ${permissions.personFields.id} for not existing rolle ${rolleId}`,
+            );
+            return Err(new EntityNotFoundError('Rolle', rolleId));
+        }
```
clauyan, line 77:
```suggestion
        const rolle: Option<Rolle<true>> = await this.rolleRepo.findById(rolleId);
        if (!rolle) {
            this.logger.error(
                `applyRollenerweiterungChangesForRolle called by ${permissions.personFields.username} - ${permissions.personFields.id} for not existing rolle ${rolleId}`,
            );
            return Err(new EntityNotFoundError('Rolle', rolleId));
        }
```

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
```diff
@@ -0,0 +1,242 @@
...
+        const removeErweiterungenPromises: Promise<{
+            serviceProviderId: string;
+            result: Result<null, DomainError>;
+        }>[] = removeErweiterungenForServiceProviderIds
+            .filter((serviceProviderId: string) => {
+                return (
+                    existingErweiterungen.findIndex(
+                        (re: Rollenerweiterung<true>) => re.serviceProviderId === serviceProviderId,
+                    ) !== -1
+                );
```
clauyan:
```suggestion
                return existingErweiterungen.some(
                    (re: Rollenerweiterung<true>) => re.serviceProviderId === serviceProviderId,
                );
```

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
```diff
@@ -0,0 +1,242 @@
...
+    }>[] {
+        const erweiterungenPromises: Promise<{
+            serviceProviderId: string;
+            result: Result<Rollenerweiterung<true>, DomainError>;
+        }>[] = addErweiterungenForServiceProviderIds
+            .filter((serviceProviderId: string) => {
+                return (
+                    existingErweiterungen.findIndex(
+                        (re: Rollenerweiterung<true>) => re.serviceProviderId === serviceProviderId,
+                    ) === -1
```
clauyan, line 144:
```suggestion
                return existingErweiterungen.every(
                    (re: Rollenerweiterung<true>) => re.serviceProviderId !== serviceProviderId,
```

src/modules/rolle/domain/apply-rollenerweiterungen-for-rolle-service.ts
```diff
@@ -0,0 +1,242 @@
...
+                    .then((result: Result<Rollenerweiterung<true>, DomainError>) => ({
+                        serviceProviderId,
+                        result,
+                    }));
+            });
+
+        return erweiterungenPromises;
+    }
+
+    private handleRemoveErweiterungen(
```
clauyan, line 192: Why does delete not check permissions but create does?

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
```diff
@@ -0,0 +1,242 @@
...
+export class ApplyRollenerweiterungWithRoleForAngeboteService {
+    public constructor(
+        private readonly logger: ClassLogger,
+        private readonly serviceProviderRepo: ServiceProviderRepo,
+        private readonly organisationRepo: OrganisationRepository,
+        private readonly rolleRepo: RolleRepo,
+        private readonly rollenerweiterungRepo: RollenerweiterungRepo,
+    ) {}
+
+    public async applyRollenerweiterungChangesWithRoleForAngebote(
```
clauyan: This function is more complicated, than it needs to be in my opinion. A lot of the consistency checks can be skipped, because the repo already handles them (via `RollenErweiterung.checkReferences` in create and a manual check for existence in delete). I suggest removing redundant checks and then checking if the two helpers are still necessary or if we can get away with one function for this.

src/modules/rolle/domain/apply-rollenerweiterung-role-angebote-service.spec.ts
```diff
@@ -0,0 +1,353 @@
...
+
+            const result: TresultType = await service.applyRollenerweiterungChangesWithRoleForAngebote(
+                orgaId,
+                rolleId,
+                body,
+                permissions,
+            );
+
+            expect(result.ok).toBe(true);
+            expect(rollenerweiterungRepo.createAuthorized).toHaveBeenCalled();
```
clauyan: I think it would make sense to assert for correct arguments here (same for delete):

```suggestion
            expect(rollenerweiterungRepo.createAuthorized).toHaveBeenCalledExactlyOnceWith(
                expect.objectContaining({
                    organisationId: orgaId,
                    rolleId,
                    serviceProviderId: serviceProviderIdAdd,
                }) as Rollenerweiterung<false>,
                permissions,
            );
```

src/modules/rolle/api/rolle.controller.integration-spec.ts
```diff
@@ -1609,4 +1613,356 @@ describe('Rolle API', () => {
...
+                    administeredBySchulstrukturknoten: organisation.id,
+                    rollenart: RollenArt.LEHR,
+                }),
+            );
+            if (rolle instanceof DomainError) {
+                throw Error();
+            }
+
+            const serviceProvider1: ServiceProvider<true> = await createAndPersistServiceProvider(em);
+            const serviceProvider2: ServiceProvider<true> = await createAndPersistServiceProvider(em);
```
clauyan, line 1634: Since DB is cleared between tests, we should probably add another angebot, so this test can actually fail if we return too much.
```suggestion
            await createAndPersistServiceProvider(em);
```

src/modules/rolle/api/apply-rollenerweiterung-service-providers.error.ts
```diff
@@ -0,0 +1,19 @@
+import { DomainError } from '../../../shared/error/index.js';
+import { MultiDomainError } from '../../../shared/error/multidomain.error.js';
+
+export class ApplyRollenerweiterungServiceProvidersError extends MultiDomainError {
```
clauyan: i18n-mapping for this error is missing, I think.

src/modules/rolle/api/apply-rollenerweiterung-for-rolle-changes.path.params.ts
```diff
@@ -0,0 +1,24 @@
...
+        required: true,
+        nullable: false,
+    })
+    public readonly rolleId!: string;
+
+    @IsString()
+    @IsUUID()
+    @IsNotEmpty()
+    @ApiProperty({
+        description: 'The spshPersonId of the person.',
```
clauyan:
```suggestion
        description: 'The organisationId of the organisation.',
```

src/modules/rolle/api/apply-rollenerweiterung-changes.body.params.ts
```diff
@@ -0,0 +1,18 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { IsUUID, IsArray } from 'class-validator';
+
+export class ApplyRollenerweiterungChangesBodyParams {
+    @IsUUID('all', { each: true })
+    @IsArray()
+    @ApiProperty({
```
clauyan, line 7: I think these should be required and maybe get a short description.

src/modules/service-provider/api/provider.controller.ts
```diff
@@ -133,7 +135,7 @@ export class ProviderController {
     }
 
     @Get()
-    @ApiOperation({ description: 'Get service-providers available for logged-in user.' })
+    @ApiOperation({ description: 'Get service-providers.' })
```
clauyan, line 138: This breaks the frontend. Either update the frontend to use the new endpoint or keep the existing endpoint the same and have the new endpoint be `GET provider/available` or similar.

src/modules/service-provider/api/find-angebote-query.params.ts
```diff
@@ -0,0 +1,41 @@
...
+        description: 'The name for the angebot.',
+        required: false,
+    })
+    public readonly searchStr?: string;
+
+    @IsOptional()
+    @IsUUID()
+    @ApiProperty({
+        description: 'The id of the organisation where the angebot should be available.',
+        required: false,
```
clauyan, line 22: Since the function returns nothing without this id, I'd make it required instead.

src/modules/service-provider/api/find-angebote-query.params.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    @IsOptional()
+    @TransformToArray()
+    @IsEnum(RollenSystemRechtEnum, { each: true })
+    @ArrayUnique()
+    @ApiProperty({
+        enum: RollenSystemRechtEnum,
+        nullable: true,
+        enumName: RollenSystemRechtEnumName,
+        required: false,
```
clauyan, line 34: Same as above.

src/modules/service-provider/domain/service-provider.service.ts
```diff
@@ -344,4 +344,28 @@ export class ServiceProviderService {
             rolle: rollen.get(rollenerweiterung.rolleId)!,
         }));
     }
+
+    public async findAllowedProvidersForRollenerweiterungAtOrga(
```
clauyan: Good function name! Can you move it up, so the public functions are at the top and then the private functions follow?

clauyan (changes requested):
I just wrote down everything that stuck out to me. I'm open to chat about any points you disagree with :)

src/modules/rolle/api/rolle.controller.ts
martkrau: I asked Caspar about this, so i have both system rights there

src/modules/rolle/api/rolle.controller.ts
martkrau: I talked with Caspar and Marvin and we liked the opportunity to have the enpoint also without an orga.

src/modules/rolle/api/rolle.controller.ts
martkrau: For this i talked with caspar that we split the services for the apply logic in one from role to angebote and one from angebot to roles, for the endpoint itself we could discuss but its based on the role so I could see it in the rolle controller

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
martkrau: Like i said above it was the decision to split them

src/modules/rolle/domain/apply-rollenerweiterungen-for-rolle-service.ts
martkrau: I dont understand this one. the rolle is fetched before in the code already.

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
martkrau: For this one i copied the logic from the other function of Caspar, so maybe we can discuss if we want to change this one.

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
martkrau: For this one i copied the logic from the other function of Caspar, so maybe we can discuss if we want to change this one.

src/modules/rolle/domain/apply-rollenerweiterungen-for-rolle-service.ts
martkrau: Also here i copied the design of the other functions from Caspar, there is no deleteAuthorized function (yet), maybe we can discuss that with Caspar.

src/modules/rolle/domain/apply-rollenerweiterungen-for-rolle-service.ts
clauyan: Sorry, the suggestion got mapped weirdly. Before it was
```typescript
const rollen: Map<string, Rolle<true>> = await this.rolleRepo.findByIds([rolleId]);
const rolle: Option<Rolle<true>> = rollen.get(rolleId);
```
and I think we can just use `await this.rolleRepo.findById(rolleId)` instead.

src/modules/rolle/domain/apply-rollenerweiterungen-role-angebote-service.ts
martkrau: Ok maybe we can also here talk with Caspar again.

src/modules/rolle/api/apply-rollenerweiterung-service-providers.error.ts
martkrau: Do we have that on multidomain errors? I also copied the error from the other service error from Caspar.

src/modules/rolle/api/apply-rollenerweiterung-changes.body.params.ts
martkrau: Why should they be required if i want only add some and dont want to remove for example?

src/modules/service-provider/api/provider.controller.ts
martkrau: This refactoring was discussed with Caspar and Phillip. I have to change the frontend yet.

src/modules/service-provider/api/find-angebote-query.params.ts
martkrau: For now the endpoint has only the codepath where the orga would be required, but in the future there will be other paths so orga can be optional I would say.

src/modules/rolle/api/apply-rollenerweiterung-changes.body.params.ts
clauyan: I thought they were optional by default and we had a mismatch between type and API-documentation, but they are not, so you don't need to specify explicitly.

I assumed empty array is okay and covers your example, but not providing anything is wrong.

src/modules/rolle/api/apply-rollenerweiterung-for-rolle-changes.path.params.ts
```diff
@@ -16,7 +16,7 @@ export class ApplyRollenerweiterungForRollePathParams {
     @IsUUID()
     @IsNotEmpty()
     @ApiProperty({
-        description: 'The spshPersonId of the person.',
+        description: 'The organisationId of the person.',
```
clauyan: 'person' means the calling admin?

src/modules/rolle/api/rolle.controller.integration-spec.ts
clauyan: Here I meant that there should be a serviceProvider without a rollenerweiterung. This test would pass if you just returned all serviceProviders from the DB (since the test-db only has the ones you create here), but maybe I'm overthinking this.

src/modules/rolle/api/rolle.controller.ts
clauyan: I agree - but then we should probably move the angebot case into the provider-controller, following the same logic. But out of scope, I think.

src/modules/rolle/api/apply-rollenerweiterung-service-providers.error.ts
clauyan: Yes, see `dbildungs-iam-server/src/modules/rolle/api/apply-rollenerweiterung-multi-exception-filter.ts`. If the error is not mapped at all, it turns into a generic 500 and we lose the content, I think.

src/modules/rolle/api/apply-rollenerweiterung-for-rolle-changes.path.params.ts
```diff
@@ -0,0 +1,24 @@
...
+        required: true,
+        nullable: false,
+    })
+    public readonly rolleId!: string;
+
+    @IsString()
+    @IsUUID()
+    @IsNotEmpty()
+    @ApiProperty({
+        description: 'The organisationIdof the organisation.',
```
clauyan:
```suggestion
        description: 'The organisationId of the organisation.',
```

src/modules/rolle/api/apply-rollenerweiterung-multi-exception-filter.ts
clauyan, line 1: `DbiamApplyRollenerweiterungMultiError` still says `rolleIds...`, which will be confusing as caller. I am thinking about a change, that doesn't affect the frontend... If we change `rolleIds` to `ids` the error depends on context and we need to update the frontend.
Maybe we can add a second array for angebote to the error? I am not entirely convinced by the idea. Maybe have a chat with the others.

clauyan (changes requested):

clauyan (approved):

PR #1254, opened 2026-07-28
SPSH-3853

YoussefBouch:
# Description

## Approval for review


src/modules/rolle/api/rolle.controller.integration-spec.ts
```diff
@@ -713,6 +713,74 @@ describe('Rolle API', () => {
...
+                        administeredBySchulstrukturknoten: org.id,
+                        merkmale: [RollenMerkmal.MPT_ROLLE],
+                    }),
+                ),
+            ]);
+
+            permissionsMock.getOrgIdsWithSystemrecht.mockResolvedValue({ all: false, orgaIds: [org.id] });
+
+            const response: Response = await request(app.getHttpServer() as App)
+                .get(`/rolle?systemrechte=ROLLEN_ERWEITERN&systemrechte=MPT_ROLLEN_VERWALTEN&organisationId=${org.id}`)
```
clauyan: You can also do it like this, which might be a bit easier to read.
```suggestion
                .get('/rolle')
                .query({
                    systemrechte: [RollenSystemRechtEnum.ROLLEN_ERWEITERN, RollenSystemRechtEnum.MPT_ROLLEN_VERWALTEN],
                    organisationId: org.id,
                } as FindRolleQueryParams)
```

src/modules/rolle/domain/rolle-find.service.ts
```diff
@@ -29,7 +29,7 @@ export class RolleFindService {
     ) {}
 
     public async findRollenAvailableForErweiterung(
-        params: FindRollenWithPermissionsParams,
+        params: FindRollenWithPermissionsParams & { includeMptRollen?: boolean },
```
clauyan: `includeMptRollen` is only based on whether the client included the systemrecht in the request. We bypass a permission check, if we pass it directly. I suggest adding `requestedWithSystemrechte` to `FindRollenWithPermissionsParams` (or an extra interface for `findRollenAvailableForErweiterung`) and adding `RollenSystemRecht.MPT_ROLLEN_VERWALTEN` to the call of `permissions.getOrgIdsWithSystemrecht`, if the SystemRecht is included.

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -95,20 +95,9 @@ export class RolleController {
         @Permissions() permissions: IPersonPermissions,
     ): Promise<PagedResponse<RolleWithServiceProvidersResponse>> {
         let rollenAndTotal: [Rolle<true>[], number];
+        const systemrechteSet: Set<RollenSystemRechtEnum> = new Set(queryParams.systemrechte ?? []);
```
clauyan, line 98: If we have a `Set` here, we could simplify the conditions in the following ifs a bit.
`systemrechteSet.size === 1 && systemrechteSet.has(...)`

clauyan (changes requested):

src/modules/rolle/domain/rolle-find.service.ts
YoussefBouch: Good catch, I thought I added a check for the permissions but seems not ... :D 

Added a check for it if the requester sends the Recht.

src/modules/rolle/api/rolle.controller.integration-spec.ts
YoussefBouch: Done

src/modules/rolle/api/rolle.controller.ts
YoussefBouch: Added size check for all other branches

src/modules/rolle/api/rolle.controller.integration-spec.ts
```diff
@@ -713,6 +714,82 @@ describe('Rolle API', () => {
...
+                .query({
+                    systemrechte: [RollenSystemRechtEnum.ROLLEN_ERWEITERN],
+                    organisationId: org.id,
+                } as FindRolleQueryParams)
+                .send();
+
+            expect(response.status).toBe(200);
+            const pagedResponse: PagedResponse<RolleWithServiceProvidersResponse> =
+                response.body as PagedResponse<RolleWithServiceProvidersResponse>;
+            // Only the non-MPT rolle should be returned
```
clauyan, line 789: You could add an assertion for this

clauyan (approved):

src/modules/rolle/api/rolle.controller.integration-spec.ts
YoussefBouch: Added assertion!

PR #1274, opened 2026-08-18
SPSH-3823: Angebotsauswahl für schulische Rollenerweiterungen einführen

MartinSchuhmacher:
# Description

https://ticketsystem.dbildungscloud.de/browse/SPSH-3823
https://github.com/dBildungsplattform/schulportal-client/pull/653

## Approval for review


src/modules/rolle/api/rolle.controller.ts
```diff
@@ -459,28 +458,23 @@ export class RolleController {
         @Query() queryParams: FindRollenerweiterungQueryParams,
         @Permissions() permissions: IPersonPermissions,
     ): Promise<ServiceProviderResponse[]> {
-        const rolleResult: Result<Rolle<true>> = await this.rolleRepo.findByIdAuthorized(params.rolleId, permissions);
-        if (!rolleResult.ok) {
+        const isAuthorized: boolean = await permissions.hasSystemrechtAtOrganisation(
```
clauyan: I think we already talked about this in our previous chat, but I don't remember the outcome. Checking permissions in the controller should be avoided. Do you think it would be possible to handle permissions with a `findByIdAuthorized`-style method in the `RolleRepo`? If time is an issue, I'd suggest creating a refactoring-ticket for this.

src/modules/service-provider/api/provider.controller.ts
```diff
@@ -173,7 +172,7 @@ export class ProviderController {
             items: serviceProviderResponses,
         };
 
-        return new PagedResponse(pagedResponse);
+        return new RawPagedResponse(pagedResponse);
```
clauyan, line 176: After some research it looks to me, as if `PagedResponse` is the default. It is mapped using a global interceptor, while `RawPagedResponse` is returned as is. The main difference is that `PagedResponse`  (after mapping) returns the paging-metadata in the **headers** and `RawPagedResponse` in its **body**. So the frontend has to consume endpoints differently, depending on which is used.

src/modules/rolle/api/rolle.controller.ts
```diff
@@ -459,28 +458,23 @@ export class RolleController {
...
-            rollenerweiterungen = await this.rollenerweiterungRepo.findManyByRolleId(params.rolleId);
-        } else {
-            rollenerweiterungen = await this.rollenerweiterungRepo.findManyByOrganisationAndRolle([
+        const rolle: Rolle<true> | null | undefined = await this.rolleRepo.findById(params.rolleId);
+        if (!rolle) {
+            throw new EntityNotFoundError('Rolle', params.rolleId);
+        }
+
+        const rollenerweiterungen: Rollenerweiterung<true>[] =
+            await this.rollenerweiterungRepo.findManyByOrganisationAndRolle([
```
clauyan: Since we want to avoid checking permissions in the controller directly, I suggest moving the check down into `RollenerweiterungRepo`. `findManyByOrganisationAndRolle` probably has more call sites with different requirements, so I would add a new function instead of extending the existing function.

src/modules/rolle/repo/rollenerweiterung.repo.integration-spec.ts
```diff
@@ -373,6 +409,45 @@ describe('RollenerweiterungRepo', () => {
...
+            await sut.create(factory.createNew(organisationB.id, targetRolle.id, serviceProviderB.id));
+            await sut.create(factory.createNew(organisationA.id, otherRolle.id, serviceProviderA.id));
+
+            const result: Array<Rollenerweiterung<true>> = await sut.findManyByRolleId(targetRolle.id);
+
+            expect(result).toHaveLength(2);
+            expect(result.every((entry: Rollenerweiterung<true>) => entry.rolleId === targetRolle.id)).toBe(true);
+        });
+
+        it('should return an empty array when no rollenerweiterungen exist for the requested rolle id', async () => {
```
clauyan: This test can't fail, since the database is cleared between tests.

src/modules/service-provider/api/provider.controller.spec.ts
```diff
@@ -585,6 +587,95 @@ describe('Provider Controller Test', () => {
...
+                expect(result.items[0]?.id).toBe(serviceProvider.id);
+            });
+        });
+
+        describe('when organisationId is missing', () => {
+            beforeEach(() => {
+                vi.clearAllMocks();
+            });
+
+            it('should skip the special branch', async () => {
```
clauyan: This label could be more clear. Maybe describe the action that should be skipped instead.

clauyan (changes requested):

clauyan (approved):

PR #1275, opened 2026-08-18
SPSH-3962 - Add Angebots Filter to rollen management view

bischofmax:
# Description

## Approval for review
- [ ] QA: In addition to review, the code has been manually tested (if manual testing is possible)
- [ ] All points were discussed with the ticket creator, support-team or product owner. The code upholds all quality guidelines from the PR-template.

PR #1276, opened 2026-08-19
SPSH-3959: Add search filter to ManageableServiceProvidersParams and update related components

bergatco:
# Description

Als Lara Landesadmin möchte ich die Einträge in der globalen Angebotsliste durch Eingabe von Angebotsnamen oder Namensbestandteilen durchsuchen können, damit ich schnell einen oder mehrere Einträge finden kann.

JIRA :
- https://ticketsystem.dbildungscloud.de/browse/SPSH-3959

Related PRs :
- https://github.com/dBildungsplattform/schulportal-client/pull/654

Deployments :
- https://spsh-3959.dev.spsh.dbildungsplattform.de/

## Approval for review


src/modules/service-provider/api/manageable-service-providers.params.ts
```diff
@@ -21,4 +21,12 @@ export class ManageableServiceProvidersParams extends PagedQueryParams {
         enumName: ServiceProviderKategorieTypName,
     })
     public readonly kategorien?: ServiceProviderKategorie[];
+
+    @IsString()
+    @IsOptional()
+    @ApiPropertyOptional({
```
bischofmax, line 27: Ergibt hier maxLength noch Sinn?

src/modules/service-provider/api/provider.controller.spec.ts
```diff
@@ -673,6 +674,30 @@ describe('Provider Controller Test', () => {
                 expect(result.items[0]?.id).toBe(matchingServiceProvider.id);
             },
         );
+
+        it('should forward the searchFilter to the service', async () => {
```
bischofmax: Ich denke es so: Wenn du 100% Coverage mit den Api Integration Tests erreichen kannst, dann würde ich keine Unit Tests für den Controller schreiben.

src/modules/service-provider/repo/service-provider.repo.integration-spec.ts
```diff
@@ -263,6 +263,122 @@ describe('ServiceProviderRepo', () => {
             expect(unterrichtCount).toEqual(1);
         });
 
+        it('should filter by searchFilter matching the name case-insensitively', async () => {
+            const [matchingServiceProvider]: [ServiceProvider<true>, ServiceProvider<true>] = await Promise.all([
+                createAndPersistServiceProvider(em, { name: 'Alpha Angebot' }),
+                createAndPersistServiceProvider(em, { name: 'Beta Angebot' }),
+            ]);
+
```
bischofmax: Braucht es hier em.clear? Hatte es das im expect die Entity mit dem Mikro Arm Cache verglichen wurde.

src/modules/service-provider/api/manageable-service-providers.params.ts
clauyan: I'm checking with PO for a possible edge case here, but assuming there is none, I suggest you copy the validation from the `create`-path.
```typescript
@IsDIN91379AEXT()
@MaxLength(50)
```

src/modules/service-provider/api/provider.controller.spec.ts
clauyan: Agreed. Prioritize integration tests that represent real usage and if you exhausted all reasonable usage via integration test, but still need more coverage, write unit tests. Doing both is obviously ideal, but since we don't have infinite time, I suggest this approach.

src/modules/service-provider/repo/service-provider.repo.ts
```diff
@@ -246,6 +246,11 @@ export class ServiceProviderRepo {
             where.kategorie = { $in: filter.kategorien };
         }
 
+        if (filter?.searchFilter) {
+            const escapedSearchFilter: string = filter.searchFilter.replace(/[\\%_]/g, '\\$&');
```
clauyan, line 250: This is a good addition, but it is the first time I see it in the codebase. Please check if MikroORM provides this functionality out of the box, if not we should consider making this available globally.
Alternatively we could tighten the validation in the controller, so these special characters don't make it this far.

clauyan (changes requested):

src/modules/service-provider/repo/service-provider.repo.ts
YoussefBouch: I think there’s a slight distinction here: this isn’t about escaping SQL input, which MikroORM already handles through parameterization. The escaping is specifically needed because `%` and `_` have special meaning inside a `LIKE`/`ILIKE` pattern.

For example, `A_B Bildungsportal` should be searchable as a literal string. Without escaping, the `_` would be interpreted as a wildcard and could match `A-B`, `A1B`, etc.

I therefore don’t think we should reject these characters in the controller, since they’re valid characters in a service provider name. If we have multiple places where user input is used in `LIKE`/`ILIKE` patterns, I’d agree that a shared helper for escaping LIKE patterns could make sense.

**That being said, I’m not sure how important this LIKE-pattern escaping actually is in our case, since service provider names are just normal strings and I’d expect %, _, etc. to be quite uncommon. If i were you, I'd just remove it.**

bergatco: > This is a good addition, but it is the first time I see it in the codebase. Please check if MikroORM provides this functionality out of the box, if not we should consider making this available globally. Alternatively we could tighten the validation in the controller, so these special characters don't make it this far.

I asked my french Colleague `Claude Opus 4.8` about it and here is the result :

--- 

`Me` :
Is there a way to let mikroorm handle the escaping of the special characters %, _, and \ in the LIKE / ILIKE statements?

`Claude` :
Short answer: **No.** MikroORM does not automatically escape LIKE/ILIKE wildcards for you.

The `$like`/`$ilike` operators only parameterize the value (protecting against SQL injection) — but `%`, `_`, and `\` inside that value are still interpreted as pattern metacharacters. There's no built-in operator or config flag that escapes them. So the manual escaping you have in `service-provider.repo.ts:250` is the correct and standard approach:

```typescript
const escapedSearchFilter: string = filter.searchFilter.replace(/[\\%_]/g, '\\$&');
where.name = { $ilike: `%${escapedSearchFilter}%` };
```

A few notes:

- This relies on PostgreSQL's **default LIKE escape character** (`\`). That's the default, so it works — but it's implicit. If you want to be explicit/robust you'd need raw SQL with `ESCAPE '\'`, which MikroORM's operator syntax doesn't let you attach. In practice the default is fine.
- The other repos in this codebase (`organisation.repository.ts:334`, `rolle.repo.ts:212`, `scope-base.ts:85`) do **not** escape at all, so a user typing `%` or `_` there gets unexpected wildcard behavior. Your version is actually the correct one.
- 
If you want to avoid duplicating the regex, the cleanest option is to extract a small shared helper (e.g. escapeLikePattern(input: string)) and reuse it across the repos, since MikroORM won't do it natively. Want me to add that helper and apply it to the search sites?

`Me` :
Is there another npm package or something like to handle the conversion for me?

`Claude` :
Not really worth it — there's no well-maintained, widely-used npm package dedicated to escaping SQL LIKE wildcards. The few that exist (e.g. tiny `sql-escape`-style packages) either target full SQL string escaping (wrong tool — MikroORM already parameterizes) or are one-line regex wrappers that add a dependency and supply-chain risk for something trivial.

MikroORM sits on top of Knex, and Knex doesn't auto-escape LIKE metacharacters either, so there's nothing in your existing stack to lean on.

The pragmatic answer: keep it in-repo as a tiny shared helper rather than pulling in a package. It's the same regex you already have, just centralized and testable:

```typescript
// src/shared/util/escape-like.ts
export function escapeLikePattern(input: string): string {
    return input.replace(/[\\%_]/g, '\\$&');
}
```

Then use it everywhere you build an `$ilike`/`$like` pattern:

```typescript
where.name = { $ilike: `%${escapeLikePattern(filter.searchFilter)}%` };
```

This gives you one place to unit-test the edge cases (`%`, `_`, `\`) and fixes the unescaped search sites in the other repos too. Want me to create the helper and wire it into the four search locations?

---

Therefore I will keep this escaping here. I would rather do an escaping here, than have the user wonder about the weird behavior of the filter. I agree that the three characters `%`, `_`, and `\` should be very seldomly used. But when the user uses these characters involuntary, he might be confused about the results. Therefore I strongly pledge for this deterministic behavior.

I also agree that this behavior should be used across all repos. @bischofmax already overtook the solution to his PR #1275 and also created a ticket to apply these change to all other repos (see [SPSH-4139](https://ticketsystem.dbildungscloud.de/browse/SPSH-4139)).

src/modules/service-provider/api/manageable-service-providers.params.ts
bergatco: Like discussed yesterday, I would add no `maxLength` for now, as this is about the search string used in the filter in the frontend and in order to enforce this limit, we would have to limit the input in this search field. And right now, I don't see a good way to inform the user about this limitation.

